PR #1, opened 2024-04-02
EB-29 Create basic hugo website layout

RMCrean:
As shown in the meeting, this PR is for the basic Hugo site layout

**Two things in particular I would like to check:**

1. I've edited the README (inside hugo) to add some extra flags to the hugo serve command. Do you agree with these flags? It seems like different people recommend different flags when developing/editing the website. 

2. Each species has been setup such that each tab will need a different page inside the given species subfolder. I think this will be okay as in the future we can write a script to create a new species folder with these files partially prefilled with yaml. Are you okay with this setup?


akochari (approved):
All looks good to me! Great job also with the search using pagefinder. I never got around to making search work in Hugo sites myself :)

RMCrean:
Great, thanks for checking!

PR #2, opened 2024-04-08
EB-22 Make script to get taxonomy info for each species 

RMCrean:
This PR does 2 main things:
- Making a python script to extract taxonomy info given a list of species.
- An example of how to include the taxonomy information from this script and display it on the Hugo site.


RMCrean:
(Merging myself as no conflicts and we are still in early stages)

PR #3, opened 2024-04-22
Improve species display on Hugo site 

RMCrean:
- (EB-37) Implement stats page on Hugo using the template provided. 
- Use Bootstrap cards to display each species, and styling of these cards, with filter by tags included. 
- Integrity checks for all scripts, and improve accessibility. 
- Setup so all data sources for a species is in a single folder (per species)


RMCrean:
(Merging myself because project still in early stages)

PR #6, opened 2024-05-03
EB-46: Add pre-commit and GH actions

RMCrean:
### pre-commit
See the .pre-commit-config.yaml file. Some generic checks alongside [markdownlint](https://github.com/DavidAnson/markdownlint) for files inside the hugo folder and [ruff](https://github.com/astral-sh/ruff) for any python files. There is a ruff.toml file which makes python files with lines longer than 120 not fail the commit. 

I've added some documentation about how to set up pre-commit in the main README. 

### GitHub actions:

As it is currently setup, no action is required to pass for a commit/merge to occur. 

4 GitHub actions added:

1. Trivy - security scanning - runs on all commits/PRs to main or develop. 
2. lighthouse - test Hugo site, for speed and accessibility etc... 
     This is only triggered if a file is changed inside the hugo folder. 
     You get a html report for each webpage checked that is available from the action logs (for 14 days).  

3. build and push docker container:
    Only triggered if a version tag is added to the commit. 

4. pre-commit with same configuration as the locally run pre-commit

### Specific Questions: 
1. Trivy - at the moment the security scanning is run on the repository just after checkout. I see in the [README of the trivy action](https://github.com/aquasecurity/trivy-action) that you can instead build a docker image and then scan that. Does that make more sense to do instead? 
2. Should we add any specific checks for bash/shell scripting? @kwentine you've wrote most of the shell scripts so far so if you have a preference here feel free to add it.

I welcome any comments about things to add/remove or improvements to any of the actions workflow files.


.github/workflows/container_reg.yml
```diff
@@ -0,0 +1,46 @@
+name: Build and push image to GitHub Container Registry
+on:
+  push:
+    tags:
+      - 'v*.*.*' # i.e. has to have a release tag to run
```
sandstromviktor, line 5: Could be nice to add a workflow dispatch here so that you can run the workflow manually, to deploy test images etc. 
see [this link](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)

sandstromviktor (commented):
Not sure if i am the one to approve, but the workflows looks good to me. I added a suggestion to add manual trigger to workflows so that you can test stuff easier.

.pre-commit-config.yaml
alfredeen, line 1: You could consider adding checks
  - id: check-merge-conflict
  - id: detect-private-key

alfredeen: Consider adding black, flake8, mypy and isort checking.

alfredeen (approved):
Made a few suggestions. Otherwise lgtm with the current branch strategy.

.github/workflows/container_reg.yml
RMCrean: Thanks, I didn't know about this option!   
I have now added a commit 4a98b6cac55128e1ab678b0c8ca0db65226e584f to add this in. And put a parameter in to control the docker tags. Does that look good? 

It seems the PR must be merged before I can test the workflow_dispatch option unfortunately (I can't see it in the actions sidebar at least).

.pre-commit-config.yaml
RMCrean: Thanks for the suggestions. I've added the: check-merge-conflict and detect-private-key options. 

I had assumed the [ruff linter](https://docs.astral.sh/ruff/linter/) had added its replacement of isort and flake8 etc.., but it was not by default active. I've updated the ruff.toml file to add in isort and flake8 and then made sure all the python scripts now meet this criteria etc...    

The [ruff formater](https://docs.astral.sh/ruff/formatter/) is an alternative to black. 

For mypy, I'm not sure about whether to enforce type checking so I wont do it at this stage unless others on the project specifically want it. 

The commit with these changes is 08d84e0a8b02cc9f4cf85f179b211257edb9235d

Thank you for checking the PR, I'm going to click resolve as you have already approved the PR but please feel free to reopen of course if you want to add something!

.github/workflows/container_reg.yml
alfredeen: Yes. I believe that the workflow with the dispatch must reside in the default branch.

RMCrean: Thanks, I will test the build once the PR is merged then and update it if needed.

kwentine (approved):
Thanks for laying this comprehensive foundation! 👍🏻

PR #9, opened 2024-05-17
WIP: Jbrowse inside Hugo Species tab

RMCrean:
Hey, I've had a go at putting JBrowse inside the browser tab of the hugo site for each species. I've had mixed success. 

I've labelled this as a draft PR for now because I'm not sure if this is the best way to set this up, but thought this would be a good format to share this with you @kwentine. 

Overview: I have it working seemingly for Clupea Harengus but other species don't seem to work. Not sure the exact reason for that. 

**P.S. I know this is Friday afternoon, so don't feel like you need to look at this till next week :)**

### Some explanation of what I've done: 

After making this new branch I ran:

```
make clean
make build
make install
```

I then used the [code in this repo](https://github.com/GMOD/jbrowse-react-app-vanillajs-demo) as a starting point as you suggested - it helped a lot! The first commit was essentially just making sure this worked (so no config file used). Initially it didn't work but after changing the version of React from 17 to 18 it worked, not sure if this is the right solution, but I made an issue in the original repo (see here: https://github.com/GMOD/jbrowse-react-app-vanillajs-demo/issues/1 ) about this, so hopefully we can figure out what is the right thing to do there.  

Then I modified the species_browser.html (full path: "hugo/layouts/species/species_browser.html") file to provide the config file as a parameter:

```
{{ if .Params.jbrowse }}
    <!-- this get the config file path from the browser.md file for each species -->
    <div id="jbrowse_app" data-config-file-path="{{ .Params.config_file }}"></div>
    <script type="module" src="/js/genomeView.js"></script>
{{ end }}
```

This will therefore provide the ".Params.config_file" value located in the header of each species markdown file (e.g. file path: hugo/content/species/clupea_harengus/browser.md) to the javascript function. 

Then inside "hugo/static/js/genomeView.js" I adapted the code from the repo to use the config.json file. This didn't work at first (data files not found errors) so I edited the config.json files to edit their local file paths. 

From for example: 
`"uri": "GCF_900700415.2_Ch_v2.0.2_genomic.gff.bgz",`
TO:
`"uri": "/data/clupea_harengus/GCF_900700415.2_Ch_v2.0.2_genomic.gff.bgz",`

This seems to work now for clupea_harengus but not for others... 

### If you want to test these changes locally you should be able to just do:
```
cd hugo 
hugo server --disableFastRender --noHTTPCache --ignoreCache
```
If you run the make commands again I guess you would need to manually edit the config file paths like I have done in this commit. 

### Working and not working example

clupea_harengus_working

Here it seems to be working fine after editing the data paths. (It took a few seconds to load once I clicked on the browser tab). 

404_for_clouded_apollo
The clouded apollo on the other hand gives this error. I can't see any issues with the file paths and if I change the paths in the config file I get an error message earlier, so perhaps it has something to do with transferring only a portion of the data? Really not sure here.


kwentine:
I think this is a great start Rory! I have adapted the JBrowse config generation to include absolute paths to the data files (to match the changes you made by hand).

However I would like to investigate a bit further before we merge: I had the impression that  when passed the `?config=path/to/config.json` URL parameter, JBrowse interpreted all paths as relative to `config.json`. I want to see if it is possible to reproduce this behavior when we get the config path from the HTML (very neat by the way!)

kwentine:
Another thought: with the tab-like interface making it very easy for the user to jump around, I think we will quickly get the remark that it is unfortunate that the browser state is lost in the process. That is of course because tab switches in fact incur page reloads.

As an alternative, I was therefore thinking: we could have a swedgene-customized JBrowse in `static/data/index.html` (the layout would be species independent though) and have it so that a click on "Browser" opens `static/data/index.html?config=data/aspergillus_nidulans/config,json` in a *new browser tab* That way the user is less likely to loose a browser setup by mistakenly clicking on "Description" or the like.

kwentine:
Yes I'm happy to continue hacking on this branch. I'll tell you when I have a proof of concept of sorts and we can move forward from there!

RMCrean:
As discussed, the changes look good, thanks. 

I added a small change to reformat the html block and VSCode was complaining about not literally assigning `{{ .Params.config_file }} ` as a string, so I've wrapped it in quotes to "fix" that. 

If it looks good to you, please feel free to merge the PR :)

PR #11, opened 2024-05-28
EB-58: Open Jbrowse in new tab

RMCrean:
Hi @kwentine, it seems to be working nicely. Please take a look and let me know what you think. 

In summary: 
 
Instead of the Browser tab for each species, a button is now shown on each tab:
image

This button includes the config in the address as planned, so you're taken to something like this currently. 
Screenshot from 2024-05-28 13-43-31

I'm not fussed on exactly we keep/how we style the JBrowse page or the exact styling of the button etc.. but that should be easy enough to change later (as you may be able to see, we can control what part of the header/footer is included in the genome-browser page by modifying `hugo/layouts/_default/baseof.html `).

Thanks!


hugo/content/species/aspergillus_nidulans/_index.md
```diff
@@ -12,6 +12,7 @@ img_attrib_text: "[EDIT]"
 img_attrib_link: [EDIT]
 
 lineage_data_path: "aspergillus_nidulans/taxonomy"
+jbrowse_config: "/data/aspergillus_nidulans/config.json"
```
kwentine, line 15: If we are consistent (and who wouldn't strive for that), `species/<latin name>/_index.md` references `/data/<latin name>/config.json` so `jbrowse_config` could be inferred. This would also avoid typos. I don't know the right way to achieve this though. Maybe using Hugo [archetypes](https://gohugo.io/content-management/archetypes/)  for species, and the [`Dir`](https://gohugo.io/methods/page/file/#dir) method ? That would be fancy, we could do `hugo new content species/<latin name>/_index.md` 😎

hugo/layouts/_default/baseof.html
```diff
@@ -4,20 +4,34 @@
...
-                          <div class="row">
-                              <div class="col-md-3 order-md-first toc small">{{- partial "toc.html" . -}}</div>
-                              <div class="col-md-9 order-md-last content" id="content">{{- block "main" . }}{{- end }}</div>
-                          </div>
-                  {{ else }}
-                          {{- block "main" . }}{{- end }}
-                  {{ end }}
-          </div>
+
+            {{ if $isGenomePage }}
```
kwentine, line 16: To avoid the multiplication of conditional guards, it would be possible to define a dedicated layout used by the browser page: `layouts/browser/baseof.html`  But that refactoring could be attempted in another PR.

hugo/layouts/partials/open_browser.html
```diff
@@ -0,0 +1,19 @@
...
+    {{ $parentPage := .Parent }}
+    {{ $configFile = $parentPage.Params.jbrowse_config | urlize }}
+{{ end }}
+{{ $browserURL := print .Site.BaseURL "genome-browser/?config=" $configFile }}
+
+<div class="row mt-3 mb-3">
+    <div class="col-12 scilife-subsection text-center">
+        <a href="{{ $browserURL }}" target="_blank">
+            <button class="btn btn-lg" title="open the genome browser in a new tab">
+                Click here to open the genome browser
```
kwentine, line 14:
```suggestion
                Browse the genome
```

hugo/layouts/partials/species_nav.html
```diff
@@ -13,14 +13,16 @@
 
     </li>
     {{ range where .CurrentSection.Pages "Section" $currentSection }}
-    <li class="nav-item">
 
-        {{ if eq $current_page .Page }}
-            <a class="nav-link active" href="{{ .RelPermalink }}"> {{ .Name }} </a>
-        {{ else }}
-            <a class="nav-link" href="{{ .RelPermalink }}"> {{ .Name }} </a>
+        {{ if ne .Name "Browser" }}
```
kwentine, line 17: Is this necessary, since `species_browser.html` was deleted ?

kwentine (commented):
I think this is the way to go 👍🏻 I made some suggestions, but none is blocking. Let me know what you think!

hugo/layouts/_default/baseof.html
RMCrean: Yes, that sounds good. I suppose we can do this once we have feedback about exactly what should be included in the browser page (header/footer wise).

hugo/content/species/aspergillus_nidulans/_index.md
RMCrean: Great idea! I have implemented it now with Hugo in commit [79c3ce3](https://github.com/ScilifelabDataCentre/swedgene/pull/11/commits/79c3ce38d31bda8ab8fb6c7983be4644bb4441ab) 

I have made a note to also do this with some other params, but will save that for another time as a bit out of scope of this PR.

hugo/layouts/partials/species_nav.html
RMCrean: Good point. I've removed it now in this commit: [79c3ce3](https://github.com/ScilifelabDataCentre/swedgene/pull/11/commits/79c3ce38d31bda8ab8fb6c7983be4644bb4441ab)

PR #19, opened 2024-07-04
EB-84: Populate "add-species" markdown pages

brinkdp:
Added content to the 'add a new genome' info page. Created sub-pages for the recommendations texts since they are long.

Also updated the CSS to improve styling. This is the main reason for the PR, but please also look at the textual content if you want/have time.

- Markdown code notation 
The file format recommendations page use quite a few code examples formatted with back-ticks. The default configuration resulted in colors that deviate from the overall styling of the web page.
Thus updated the CSS with `code {}` and `pre {}` to use `--scilife-light-grey` for backgrounds, and black font. 
Had to use `!important` to override other rules to get it to display. Perhaps this is not best-practice? Please make changes (or revert) as you see fit!

- h3 and h4
In `_index.md` and `recommendations_for_making_data_public.md`, I used h4 instead of h3 to make it look better. But then I recall that we talked at some point about how this might break the accessibility rules for screen readers? What are your thoughts here?

### DIFF
--- a/hugo/content/add_genome/recommendations_for_file_formats.md
+++ b/hugo/content/add_genome/recommendations_for_file_formats.md
@@ -2,15 +2,17 @@
 title: Recommendations for file formats
 ---
 
+## Recommendations for file formats
+
 This page contains recommendations for the file formats that can be used for displaying data on the Genome Portal and is intended as a support for researchers that want to add a new genome to the site. The text assumes some previous knowledge about bioinformatics file formats and familiarity with command line tools. The Genome Portal staff will also be happy to discuss choice of file formats and give advice on file format conversion. Please see the <a href="/contact">Contact page</a> for how to get in touch with us.
 
 The Genome Portal uses JBrowse 2 to display the datasets. JBrowse 2 <a href="https://jbrowse.org/jb2/features/#supported-data-formats">supports several formats</a> that are commonly used in genomics and transcriptomics and these can thus technically be displayed in the genome portal. However, at the moment we do **NOT display BAM files** in the genome portal since they can be quite big and might affect performance.
 
 - We still encourage research groups to make their BAM files publicly available, since that enables users to load the data as a local custom tracks in their JBrowse 2 instance.
 
-Readers that have looked at the <a href="https://jbrowse.org/jb2/features/#supported-data-formats">list of file formats supported by JBrowse</a> might have noticed the mention of index files. Indexing of files is taken care of on the server side of the Genome Portal, and therefore index files do not need to be supplied in the process of adding a new species to the portal. For advanced users, a description of how indexing can be performed is available in the [Additional Resources](#additional-resources) section below.
+Readers that have looked at the list of file formats supported by JBrowse might have noticed the mention of index files. Indexing of files is taken care of on the server side of the Genome Portal, and therefore index files do not need to be supplied in the process of adding a new species to the portal. For advanced users, a description of how indexing can be performed is available in the [Additional Resources](#additional-resources) section below.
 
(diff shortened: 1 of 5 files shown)


hugo/content/add_genome/recommendations_for_file_formats.md
```diff
@@ -0,0 +1,155 @@
...
+
+#### GFF (and GTF)
+
+One of the most common formats for storing annotations for genomic features is GFF (General Feature Format). NCBI for instance uses this for Protein Coding Genes. JBrowse 2 specifically supports the GFF3 version, and older GFF formats thus need to be converted to GFF3 as described below.
+
+GTF is another format for storing annotations that is very similar format to GFF. However, JBrowse 2 does support GTF files and they therefore also need to be converted to GFF3.
+
+There are several bioinformatics tools that can convert GTF and older GFF versions to GFF3. We recommend using <a href="https://agat.readthedocs.io">AGAT</a>. This is a command line toolkit designed by the NBIS bioinformatics platform at SciLifeLab. The AGAT script `agat_convert_sp_gxf2gxf.pl` is used to achieve conversions from GFF and GTF to GFF3.
+
+```text
```
RMCrean, line 53: This reason why !important is required is because of "text" being added to each code block. I think the easiest work around would be to remove "text" from the multi-line code blocks, and then !important can be removed. 

As "text" doesn't add any benefit as far as I can see?

hugo/static/css/styles.css
```diff
@@ -288,3 +288,17 @@ iframe#matoOpOut {
...
+  color: black;  /* A shade of gray #6e6e6e;*/
+  background-color: var(--scilife-light-grey);
+}
+
+pre {
+  color: black !important;
+  background-color: var(--scilife-light-grey) !important; 
+  -moz-tab-size: 4;
+  -o-tab-size: 4;
+  tab-size: 4;
```
RMCrean, line 303: Could add the following params for styling, feel free to choose none/some/edit of course:
```
  border: 5px solid var(--scilife-medium-grey);
  border-radius: 10px; 
  line-height: 1.5;
  padding: 5px;
```

RMCrean (approved):
As said on slack, It looks really nice, thanks :). 

I added a comment about the !important thing explaining its origin/how to avoid using it. I don't think it would be bad to use !important as this wont be a huge website, so feel free to use it here if you want, but I think we can quite easily avoid it at the same time. 

For the h3 and h4 thing. The accessibility requirement is that you should ideally go down sequentially on a page from h1 to h2 to h3 to h4 etc... So it shouldn't jump from h2 to h4 for example. (Each Page already has a h1 which is the "Genome Portal" banner). The complaint the [WAVE](https://wave.webaim.org/extension/) makes is skipping h2 on the page "Recommendations for file formats". One solution to that could be adding a line to the top of the markdown with "##Recommendations for file formats" for example. I'd recommend trying out the WAVE extension though (linked above) if you want to see what it likes/does not like. 

I think you have it from here but feel free to ask if you want me to look at something else or want a second opinion.

hugo/content/add_genome/recommendations_for_file_formats.md
brinkdp: Nice catch, thanks! This explains a lot of the quirks I ran into earlier today :)

The reason I added "text" was that markdownlint-fix was complaining:
`MD040/fenced-code-language Fenced code blocks should have a language specified [Context: "```"]`.

I agree that it is better to avoid !imporant (especially since the fix was so simple). Will also update .markdownlint.yaml so that it skips the "fenced-code-language" check.

PR #20, opened 2024-07-10
EB-91: Put docker files into a separate directory 

RMCrean:
I've now moved the Dockerfiles into a separate repo and updated the GH actions workflows to recognize this. 

To test these packages/images get correctly made, I've temporarily made it so that pushes to this branch trigger the build action. Before merging this will removed from the actions files. 
e.g. from `.github/workflows/hugo-container.yml`
```
name: Build and push Hugo site image to GitHub Container Registry
on:
  push:
    branches: # temporary, will be removed before merging to main. TODO 
      - docker-dir
    tags:
      - 'v*.*.*' # i.e. has to have a release tag to run 
```

As can be seen from the packages list for this repo, these 2 new images are successfully made. https://github.com/orgs/ScilifelabDataCentre/packages?repo_name=swedgene  The image/package named "swedgene" is what the image for the hugo website was called before. 

### Package Names

- I renamed the hugo image to "hugo-site" from "swedgene" to avoid confusion. 
- I named the other image "data-builder"

(These names are specified in the workflow files - by the `image `parameter.)

Of course I'm happy to change these names if you feel there is something more sensible to use. 

I've pulled both images and tested them locally myself and they work fine (for me at least). 

### To test the hugo-site image I did as follows: 
```
docker pull ghcr.io/scilifelabdatacentre/hugo-site:docker-dir
docker run -p 8080:8080 ghcr.io/scilifelabdatacentre/hugo-site:docker-dir
```

### For the data-builder image I did as follows: 
```
docker pull ghcr.io/scilifelabdatacentre/data-builder:docker-dir
./scripts/dockermake.sh DOCKER_IMAGE="ghcr.io/scilifelabdatacentre/data-builder:docker-dir"
```

So for this one I added an extra parameter to the `dockermake.sh` file to specify the docker image. In the shell script the default image I chose ("ghcr.io/scilifelabdatacentre/data-builder:latest") would only work once the main branch has run the GH action. That seems like the right choice to me, but say if you disagree. 

I've added a small amount of info to the README about this, but feel free to extend it. I think the README needs some work and some reorganization (we could have a separate Contributing.md file for example), but that might be best saved for later to keep this PR focused etc..


kwentine (approved):
Looks great! I just parametrized the docker-related scripts using environment variables rather than command line arguments.

And a good review should always include a nitpick: how about we name the folder `docker` instead of `dockerfiles` ? 🙃  Feel free to disregard if not tempted by the change.

RMCrean:
Awesome, thanks for checking and adding the edits to the shell script! 
I've just made 2 commits to finish things off:  

1. To rename the folder to "docker" as suggested. 
2. Now that the images were built fine with the commit above (i.e. I didn't miss any file paths to update), I've removed the building of these images on pushes. So now they only can occur on the main branch for now from us manually running the action or creating a version etc...  

Merging now, thanks again :)

PR #23, opened 2024-07-17
Rename Docker images

kwentine:
Prepend `swg-` to our public Docker image names to avoid naming conflicts.
Also included the script used to perform the renaming , if that is something we get into the habit of doing 🙄


RMCrean (approved):
Thanks for doing this, looks good to me. I've just renamed the mentions of the image names in the README and committed/pushed that now.

kwentine:
Thank you! 👍🏻

PR #24, opened 2024-07-17
Make species data directories readable to the world

kwentine:
This tiny pull request ensures that species directories created by make in `DATA_DIR` are "readable by everyone". In a Kubernetes context, this means that the `nginx` user will be able to read (and hence serve) the JBrowse configuration files.

The proposed modification was necessary because the `--create-dirs` option of `curl`, creates intermediate directories with mode 0750, which makes them unreadable for a user that is not the owner or a member of the group owning the file.


RMCrean (approved):
Looks good to me, thanks for looking into this.

PR #25, opened 2024-07-18
Testing environment for genomic data build

kwentine:
This PR sets up a testing environment for our data builds. 

- Test configurations are stored in  `tests/configs`
- Test data is stored in `tests/fixtures`

Then `scripts/dockermake.sh --test <make target>` will:
- Start a web server making test data available at the `http://fixtrures` URL
- Run `make <recipe>`

The generated data will be in `tests/data` and `test/public` (if `make install` is being tested)

Feel free to commit test data to the PR if you are inspired, but it can also be added later as we add support for data formats or other `make` features.

Note: @brinkdp may have to prefix the test command with `SWG_USER=0` to avoid filsystem permission issues


scripts/dockermake.sh
```diff
@@ -4,24 +4,79 @@
...
+
+_DOCKER_ARGS=()
+
+# Run make in Docker container using host Makefile, with data,
+# configuration and installation directories mounted.
+docker_make() {
+    _DOCKER_WORKDIR=/swedgene
+    # Make sure writable directories exist on the host
+    mkdir -p "$SWG_DATA_DIR" "$SWG_INSTALL_DIR"
+    docker run --rm "${_DOCKER_ARGS[@]}" \
```
brinkdp, line 28: Related to the ARM64 vs AMD64 issue, I mentioned the other day,
`scripts/dockermake.sh --test` eventually runs into this error on my Macbook:

`docker: image with reference ghcr.io/scilifelabdatacentre/swg-data-builder:main was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64.`

I am able to I get around this error by adding `--platform=linux/amd64 \` here as an argument for `docker run`.
The image still gets flagged with the !AMD64 warning in my docker client, but I am now able to run `scripts/dockermake.sh --test` commands.

I'll continue testing with my modified `dockermake.sh` for now, but will not commit the change. Maybe there is a better solution than to hardcode the platform into the script?

brinkdp:
Thanks for this! It's a great idea to have a testing environment with synthetic dataset with a small footprint.

Continuing after adding `--platform=linux/amd64 \`  as per earlier comment,  `scripts/dockermake.sh --test build` now executes without error. 

Here I had to pause for a minute to enjoy a good hearty laugh triggered by the name of the test species. 
It was beautiful <3 

Anyway. It builds as expected, and config.json looks good. Specifying `SWG_USER=0` was not needed for me for this command.

Testing `install` gave this, however. Looks like one of the errors I encountered in PR17.

```
❯ scripts/dockermake.sh --test install
Creating test network swg-test-net
13deb3d0eb24ce9cd400c39c125d80e9b0a13eeaaa75d8e1bbbe2e1c2d6e4748
Starting test server...
49b9b1f50ca0d0dc81f2078d0adf74d2a462ca8cd3be751822bc893ccba2035a
cp: target directory 'hugo/static': No such file or directory
make: *** [Makefile:97: install] Error 1
Cleaning up...
```

Could it be related to me not having the correct image mounted? I tried `docker pull ghcr.io/scilifelabdatacentre/swg-hugo-site:main` followed by `scripts/dockermake.sh --test install`, but no avail.

To check that the test data displays as intended, I instead copied the files created in `test/data` to a local JBrowse web install and ran `npx serve .`. Looks good. I will add another test track that utilizes the alias files so that we cover that use-case too.

scripts/dockermake.sh
kwentine: Until we have found a solution to perhaps build images for multiple architectures as part of our CI, I would suggest that you build your images locally. For example:
```
./scripts/dockerbuild.sh data
SWG_TAG=local ./scripts/dockermake.sh --test debug
```
The first command will build the `ghcr.io/scilifelabdatacentre/swg-data-builder:local` image on your computer. Specifying `SWG_TAG=local` will instruct `dockermake.sh` to use that one instead of the default `ghcr.io/scilifelabdatacentre/swg-data-builder:main` that is remotely built.

kwentine:
Thanks again for very helpful feedback!

> Here I had to pause for a minute to enjoy a good hearty laugh triggered by the name of the test species. It was beautiful <3

Haha I am so pleased that you noticed ❤️ 
 
> Testing `install` gave this, however. Looks like one of the errors I encountered in PR17.

You found a bug that should be fixed by 58e2a57

>  I will add another test track that utilizes the alias files so that we cover that use-case too.

That is exactly what I hoped you'd start doing: adding samples and configuration files that correspond to use cases you encountered in real `config`, but also examples of things you'd like to support. This is a playground, so anything goes!

scripts/dockermake.sh
kwentine: Also mentioning the `SWG_DEFAULT_USER=1` environment variable you can pass to `dockerbuild.sh` if you run into user issues. This will use the default user id `1000` when building the image instead of trying to use the user ID you have on your machine. It's a bit messy still, sorry.

brinkdp: Excellent, thank you so much for this advice! `SWG_DEFAULT_USER=1 scripts/dockerbuild.sh` solved the previous GID issue I had. The local image is installed (and, as expected, lacks the !AMD64 flag). I then rolled  back the `--platform` arg that I had been playing with. Using `SWG_TAG=local ./scripts/dockermake.sh --test build` works without a problem. Will continue to test other make recipes and report back!

RMCrean:
Sorry, I'm a little late on the replies, looks like you guys have made some really nice progress. Also, yes, the name choice of Tiny Herb was perfect, glad it has made it's way back into the repo where it truly belongs!  

Both: `./scripts/dockermake.sh --test build ` .`/scripts/dockermake.sh --test install ` worked for me directly (from the latest commits). 

One thing I'm a little confused about is how the testing/confirmation part is done? Is the idea that we should then open these tracks up in a local JBrowse instance and confirm that it works as expected? Or will some kind of assertion tests be made to check e.g. the config.json produced is as expected? Or something else?

RMCrean (approved):
Thanks for these changes. I can confirm the --test flagged build and install commands work for me on my end. 

I think jut needed to add the flag --test is a very elegant solution. :)

kwentine:
Thanks for your feedback! It's still very much tinkering, but I hope we can use this as a stepping stone in the future for proper CI testing. In the meantime, for tight "hack-test-repeat" loops, I also mention the `-k` flag that will keep the test containers up and thus avoid re-creating them on each run.

PR #27, opened 2024-07-23
Enhanced installation recipe

kwentine:
- `make` now only installs (_ie_ copies) new files, or files that have changed since last install
-  `make` picks up `SWG_*` variables from the environment
- `scripts/dockermake.sh` leverages the previous point to dynamically build the flags to pass to `docker run` 

Minor bug fixes include:
- tell `curl` to fails when the HTTP responses have a faulty status code (e.g. 404) instead of putting garbage in `.downloads` creating errors further down the line.
- use `patsubst` rather than `subst` to guarantee prefix-substitutions


RMCrean (approved):
Thanks, the changes look good and I can confirm make build/install still works for me. I'll start reviewing the next PR shortly too :)

PR #28, opened 2024-07-23
Support zipped GFF files

kwentine:
This PR adds support for remotely hosted GFF files with the `.zip` extension (an example of which can be found in _Linum tenue_)

The solution is ad-hoc: to create `genes.gff.bgz` look at any file in `DOWNLOAD_TARGETS` that matches `genes.gff`; that can turn out to be `genes.gff.zip` or `genes.gff.gz`. 

It works partly because `zcat` handles both `.gz` and `.zip` identically.

I have some ideas to push this work further, so that we can support any compression format handled by `zcat` for all remote files (not only GFF). But that will be for another PR.

Note: this PR will be merged in main after #27 . The target is currently set to #27 to make the diff more apparent.


RMCrean (approved):
Thanks I can confirm it works for me. 
I like the addition of the tests for zipped files and can see them being used nicely when I run with the --test flag :)

PR #29, opened 2024-07-24
Generalized compression format support

kwentine:
Building on #28, support an extensible set of compression formats for both FASTA and GFF. In particular, no compression at all is allowed. The only limitation is that compression formats be supported by `zcat`.

Highlights:
- The `std_extension` function has been rewritten to ~~use a lot of interesting bash mechanisms~~ be more robust and easily customized
- Files that are not compressed saved with the `.nozip`™ extension
- `bgzip` compression is handled by a single `make` recipe
- `scripts/filter.sh` allows arbitrary pre-gzip transformations to be applied to files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,10 +155,6 @@ $(FASTA_INDICES): %.fai: %
 $(GFF_INDICES): %.tbi: %
 	@$(SHELL) scripts/index_gff.sh $<
 
-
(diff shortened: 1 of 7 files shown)


RMCrean (approved):
LGTM, thanks. 

P.S. I enjoyed the Gandalf commit message reference :)

PR #31, opened 2024-08-01
Avoid caching download targets in targets.mk

kwentine:
No longer resort to the`targets.mk` intermediary makefile to define the `DOWNLOAD_TARGETS` variable. Instead, assign the variable afresh from a shell command every time `make` is invoked. 

This ensures that the `DOWNLOAD_TARGETS` always reflect the state of the configuration when make is invoked, also taking into consideration  `SPECIES=rosa_rosae_rosam` restrictions passed on the command line. 

The `targets.mk` was only rebuilt if a configuration file was added or modified. Not if a configuration was removed. And if `targets.mk` was generated when `make` was called without `SPECIES` argument, it would interfere with a subsequent call made with `SPECIES=foo`. So instead of a more elaborate cache invalidation logic, decided to drop the complexity altogether and incur the microscopic overhead of calling `make_download_targets` on each `make` run.

`make_download_targets` got a nice refactoring in the process and I believe is now more readable.

Finally, an important modification was to use simply expanded variables in the `Makefile` (using the `:=` assignment), making sure the variable definitions are evaluated once and immediately (rather than every time the variable is used)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ SPECIES=$(SPECIES:%:{%})
 CONFIGS = $(shell find $(CONFIG_DIR)/$(SPECIES) -type f -name 'config.yml')
 JBROWSE_CONFIGS = $(patsubst $(CONFIG_DIR)/%,$(DATA_DIR)/%,$(CONFIGS:.yml=.json))
 
-# Defines the DOWNLOAD_TARGETS variable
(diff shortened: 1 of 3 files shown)


RMCrean (approved):
LGTM :) 

P.S. That was cool to learn about the difference between `:=` and `= `in a makefile.

PR #34, opened 2024-08-16
Initial JBrowse configurations

kwentine:
This PR allows to specify initial JBrowse configurations, that will be used as a starting point to generate de fully-fledged `config.json`.

To add an initial JBrowse configuration, create `config.json` alongside `config.yaml` in the species configuration directory.

The final `config.json` will include the elements in the initial configuration, plus all those specified in `config.yaml` However if a track or assembly is present both in the initial configuration and `config.yaml`, the latter will take precedence (no fine-grained merge is performed). 

Intended use cases:
- Add initial sessions for each species
- Add tracks that require configuration more advanced than what `jbrowse add-track` can handle (BED files with column names and score column for example)
- Manually add hosted track that we don't want to download, but that would perplex `add-track` because of opaque URLS
  - Example: `.bed` file hosted at `figshare/12345`


brinkdp (approved):
This is an incredibly useful feature, thanks for this! 

I started out by testing with the raw url for the mito_krill BED file to simulate an externally hosted, non-downloaded, file. I set the track up in `config/mito_krill/config.json` and ran `dockermake --test build` followed by a local `npx seve` instance. The BED track displayed as expected, very nice!

Then, inspired by your minimal example for *L. trigynum* config.json in 7d46a34efd1eacf32a7cc7fe564c1eb517d7787a, I took a deep-dive and started playing with how to configure defaultSession. It was a very educational experience, much thanks to how fast it is to work with the testing environment. After some trial-and-error, I found settings for mito_krill's initial config that makes the JBrowse instance skip directly to the linear view, turn on the track selector menu and BED track, and set the zoom level. There are likely superfluous lines in there that perhaps could be stripped to improve readability, but overall I think this could be a good starting point for manually setting a defaultSession. 

(Also, there was a conflict based on `tests/config/mito_krill/config.yml` that I resolved, and the result was merging `main` into this branch. If that is an undesired action, please feel free to rollback my last commit!)

kwentine:
I'm so happy that you could engage so actively with the feature and test some of the use cases. 

Along with supplying default sessions (which we can maybe dedicate a branch to), it opens the possibility for quite a lot of [tweaking](https://jbrowse.org/jb2/docs/config_guides/theme/) (color palette and logo for example). 

Merging `main` periodically into a feature branch to keep up to date is a practice I like to follow as well when, like in this case:
- more than one person is involved
- there are non trivial conflicts
 The alternative (when one works alone and fancies a linear git history graph) is to rebase the feature branch on `main` and force-push.

brinkdp:
Sounds good to dedicate a branch to continue investigating defaultSessions. There are indeed many interesting configuration points to try.

PR #35, opened 2024-08-16
Suport GTF files

kwentine:
Allow to browse plain GTF files as supported by the `GtfAdapter`.

- The remote files
  - Must have the `.gtf` extension 
  - Can be compressed using `.zip` or `.gz`.

The GTF files are downloaded and decompressed by the build process, and handled by `add-track`.

Note: the linum trigynum GTF was successfully configured but did not display anything, but it could be an aliasing issue.


brinkdp (approved):
Thanks, LGTM! 

It was interesting to learn that the GtfAdapter plugin does not require an index (and cannot use them, if I understand correctly?). I am curious to do some testing later on to see whether the lack of indexing will introduce any lag in the display? Perhaps it will not be noticeable? 

The _L. trigynum_ data should display correctly now that the new alias file is in place (well, at least after the branch has been merged with main, since the alias url temporarily calls for a raw file that is yet to get its final url. I didn't foresee that when I did the commit. Sorry for the inconvenience!). There are some quirks with this assembly that made me change from the ENA version to the NCBI version for now. I also noticed that the GTF only contains annotations for the chromosomes (chr_1 to chr_10) but not for the unplaced scaffolds (chr_11 to chr_15). Thus, there will be nothing to display for the latter five scaffolds.

kwentine:
Thanks for your feedback and contributions!

> It was interesting to learn that the GtfAdapter plugin does not require an index

I agreee this has a "too good to be true" feeling about it. Although `GtfAdapter` is not a plugin (it is built-in, stands on par with `GffTabixAdapter` as far as JBrowse internals are concerned) it is not particularly emphasized in the documentation, and the performance shortcomings that you suspect may be a reason. 

So while we can support GTF on our side for Trigynum's sake, maybe we should not encourage it in our guidelines. But that should be discussed together at a team meeting perhaps. 

> Sorry for the inconvenience

Oh no don't be,  I thought is was a nice hack to put a Github URL in there :)

PR #36, opened 2024-08-22
Configure assembly display name

kwentine:
This PR adds support for a `displayName` key in the `assembly` configuration section. 

For illustrative purposes I just set the property for the herring, unsure about what convention to use (latin names or english equivalents) 

The display name appears in the initial assembly selection drop-down in JBrowse. However, the header of the default browser view unfortunately still uses the assembly `name`.

We will likely need to provide default sessions using #34, where the view display name can be configured.

### DIFF
--- a/config/clupea_harengus/config.yml
+++ b/config/clupea_harengus/config.yml
@@ -1,6 +1,7 @@
 organism: "Clupea harengus"
 assembly:
   name: Ch_v2.0.2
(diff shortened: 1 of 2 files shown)


brinkdp (approved):
Thanks, looks good to me! The case of the view name aside, displayName seem to work as intended. I agree that configuring default sessions is probably the next step forward here.

PR #37, opened 2024-08-23
Support for BED files

kwentine:
Support genomic data files in the BED format

- Assumes that the remote files are named using the `.bed` extension
- The following steps are taken
  - download local copy of remote file 
  - compress using `bgzip`
  - sort (see [discussion](https://scilifelab.atlassian.net/browse/EB-67?focusedCommentId=15378))
  -  index using `tabix -p bed`

To test the build (thanks a lot @brinkdp for the nice fixtures):
```sh
./scripts/dockermake --test SPECIES=mito_krill 
```

To quickly inspect the results in JBrowse
```sh
cp -r tests/fixtures/public/mito_krill hugo/static/data
scripts/dockerserve --dev
``` 
And visit http://localhost:8080/genome-browser/?config=/data/mito_krill/config.json


brinkdp (approved):
Works well, thanks for this! 

Currently, running the `makefile` will result in two copies of the mito_krill BED track. I was surprised at first, but realized that this is the expected behavior after the merge (b3a0541): one is the external non-downloaded and non-indexed track specified in the initial `config.json` (set to be turned on by default), and the other is the downloaded and indexed track specified in `config.yml`. I think the former could be a useful reference point for future testing purposes, so instead of removing it from the initial config, I updated both config files with new track names that reflect the difference between the two "copies" of the BED track.

RMCrean (approved):

kwentine:
> Currently, running the `makefile` will result in two copies of the mito_krill BED track.

Thanks for pointing this out. 

JBrowse identifies unique tracks using the `trackId` key, which must be unique throughout the built `config.json`. We don't specify the track IDs in `config.yaml`, so JBrowse derives them from the file name.

In your case, the ID specified in `config.json` and the one inferred by JBrowse based on `config.yaml` were different, so two tracks resulted. Would they have been the same, the `config.yaml` track would have taken precedence and overwritten the initial homonym.

brinkdp:
Interesting, thanks for the explanation. I see, the cause for this was the hard-coded track ID used for testing of enabling tracks in defaultSession.

PR #39, opened 2024-08-27
Quick peek at a data folder in JBrowse

kwentine:
Suppose you have freshly assembled data files and assorted `config.json` in `path/to/my/tiny/herb`

Then:
```sh
scripts/browse path/to/my/tiny_herb
```
Open http://localhost:3000, and voilà :)

That is, if the image exists, which you can ensure by:
```sh
scripts/browse --build
```

I'm not sure you'll find it useful, but in doubt I open this PR :)

### DIFF
--- a/scripts/browse
+++ b/scripts/browse
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# View a data directory in JBrowse
+main() {
+    if [[ "$1" == "--build" ]]; then
+	1>&2 echo "Building browser image jb2:latest"
+	docker build -t jb2 docker/jbrowse.dockerfile
+    else
+	data_dir="$(realpath $1)"
+	name=jb2-"${data_dir##*/}"
+
+	if ! docker run -d --rm --name=${name} \
+	       -v "$(realpath $1)":/app/data \
+	       -p 3000:3000 \
+	       jb2; then
+	    1>&2 echo "Error running docker"
+	fi
+	1>&2 cat <<EOF
+Browsing data directory $1 on port 3000.
+To stop and remove container:
+    docker stop ${name}
+EOF
+    fi
+}
+
+main "$@"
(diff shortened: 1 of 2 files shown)


scripts/browse
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# View a data directory in JBrowse
+main() {
+    if [[ "$1" == "--build" ]]; then
+	1>&2 echo "Building browser image jb2:latest"
+	docker build -t jb2 docker/jbrowse.dockerfile
```
brinkdp, line 7: When trying `SWG_DEFAULT_USER=1 scripts/browse --build`, I am getting this error: 
```
Building browser image jb2:latest
[+] Building 0.0s (0/0)                                                                                               docker:desktop-linux
ERROR: unable to prepare context: path "docker/jbrowse.dockerfile" not found
```
However, the build works if I change L7 to:
`docker build -t jb2 -f docker/jbrowse.dockerfile .`

I don't know if this is another of case macOS being macOS? 😄

brinkdp (commented):
Looks like a handy feature. So the difference from `dockerserve --dev` is that `browse` boots straight into the JBrowse instance without going through the hugo pages, right? I can see the use-case for that :smile:

scripts/browse
kwentine: Thanks for spotting this, fixed in e438d02 : I had omitted a dot.

For this build, ``SWG_DEFAULT_USER=1` should not be necessary.

kwentine:
>  So the difference from `dockerserve --dev`

As an example, the workflow you used  in #43: 
```sh
./scripts/dockerserve --dev
./scripts/dockermake --test SPECIES=linum_popgen
cp -r tests/data/linum_popgen hugo/static/data
```
becomes:
```sh
./scripts/dockermake --test SPECIES=linum_popgen
./scripts/browse tests/data/linum_popgen
```

PR #42, opened 2024-08-29
EB-117: Handle if user goes to /genome-browser page 

RMCrean:
This PR is designed to handle what happens if a user goes to the /genome-browser page without a config file attached as a search parameter (/genome-browser/?config=/data/parnassius_mnemosyne/config.json). Currently if you go to just "/genome-browser" you are provided with a "welcome to JBrowse message". 

Now if the user tries to go the /genome-browser page or includes a config file that doesn't exist (see e.g. search param used in the image) the following will be displayed:

Screenshot from 2024-08-29 09-29-55

I think this covers what we need to cover, but please say if you think there might be a better way to handle it. I also hope that the error text is descriptive and clear, feel free to edit/suggest some changes there too.


brinkdp (approved):
Good idea! The default JBrowse web install screen that otherwise would show up could indeed be confusing for users. The message text you wrote looks good to me: it's clear and invites action.

kwentine:
Very clean implementation, I like it a lot :)

kwentine (approved):

PR #43, opened 2024-08-29
Test fixtures for GWAS plugin

brinkdp:
We have gotten confirmation that we will soon have new population genomics data incoming. I took this as encouragement to continue the investigation of the GWAS plugin started by @Kwentine in [EB-56](https://scilifelab.atlassian.net/browse/EB-56?atlOrigin=eyJpIjoiN2UwY2U1MWJiZmVlNDc5MThmNDI4MzRmNzFiNjk2YTgiLCJwIjoiaiJ9). Big thanks for the previous progress on this! I mostly just tweaked the config and prepared new test data.

EDIT: one week later, the data has now been made public. Findings from tests of a representative file from the dataset are presented in the next post below.

--- 
The initial commit include fixtures with BED-like data and an updated version of the EB-56 config. The result is a working Manhattan plot. 

The dummy data is based on the structure of an incoming dataset I got to preview. Since that data is yet to be made public, the fixture data contains a randomized score column with a distribution based on the min-max values of the original file with some outliers sprinkled in for effect. 

To test (inspired by PR37):
```
./scripts/dockermake --test SPECIES=gwas_testing_ground
./scripts/dockerserve --dev
cp tests/fixtures/gwas_testing_ground/gwas_randomized_dummy_data.txt.bgz* hugo/static/data
cp -r tests/data/gwas_testing_ground hugo/static/data
```
And then open: [http://localhost:8080/genome-browser/?config=/data/gwas_testing_ground/config.json](http://localhost:8080/genome-browser/?config=/data/gwas_testing_ground/config.json)

Some observations:

- First I attempted to load the plain text file using `bedAdapter`, but this only resulted in an empty plot, which is similar to the result Quentin previously got with the other test data.

- Switched to `BedTabixAdapter` like in the [example in the GWAS plugin repo](https://github.com/cmdcolin/jbrowse-plugin-gwas?tab=readme-ov-file#example-config). Realized/remembered that tabix only accepts tab-delimited data ("What's in a name?" :smile: ). Reformatted with `awk '{$1=$1}1' OFS='\t' data.txt > data.tab_delim.txt`. BED sort, bgzip, tabix. The data now displayed! 

- Despite that the official specification for the BED format states that it is whitespace-delimited, tab-delimited BEDs are commonly used and are likely the easiest way to ensure compatibility with JBrowse 2. From now on, we will ask submitting researchers to tab-delimit their BED files

- BED-like files can have headers: first line need to start with # and be tab-separated. The column titles can be used to define e.g. the `scoreColumn` for the plot.

- Take-home message: ask researchers to tab-delimit their BED files.

Some things left to consider:

- Can BED-like files be used for other visualizations? Manhattan plots will not be suitable for all population genomics datasets. Histograms have already been requested.

- Should the GWAS plugin be installed in the repo instead of called remotely?

- Will the  `bedAdapter` config work with the tab-delimited data?

- The FASTA has a file size of 2 Mb despite being truncated. The config can be altered to call on the remote data instead, but a refNameAlias first need to be prepared. 

I put this a a draft PR for now because I anticipate that the incoming dataset might have additional data types that need to be tested.


kwentine:
@brinkdp First of all I must say that I am impressed by your clever workarounds, and always very clear summary. 

For complex configurations, such as the Manhattan plot, I think we should directly craft `config.json`, partly because the `add-track` CLI does not allow to specify `adapter` options such as `scoreColumn`. Other track-level advanced options could be passed to `add-track` as a JSON string, but if we are to write JSON I reckon we might as well edit `config.json` directly.

That being said, the backend *should still prepare the data* for tracks manually added to `config.json` So I propose that we add an optional `skipAddTrack`  key to `config.yaml` that, when set to `true`, would instruct `make build` to generate track files as usual, but skip the `add-track` call.

What do you think ?

scripts/filter
```diff
@@ -18,7 +18,7 @@ main () {
 	fi
     elif [[ "$1" =~ ${bed_pattern} ]]; then
 	echo "Sorting BED file $1" >&2
-	if ! grep -v "^#" | sort -k1,1 -k2,2n; then
+	if ! sort -k1,1 -k2,2n; then
```
kwentine, line 21: I would be a bit cautious here: since the header also has a column structure, it could get mingled among regular rows by the sort, and perplex `tabix`. It may be a coincidence that it works here. 

I would maybe consider something like:
```sh
{
  # Save stdin to temporary file
  tmp=/tmp/$$-bed-stdin 
  cat > "$tmp"
  # Output the various headers first
  grep -e '#'  -e track -e browser $tmp
  # Output the sorted data lines 
  grep -v '#' $tmp | sort -k1n; rm $tmp 
}
```

brinkdp:
You raise good points; a `skipAddTrack` key in combination with manual crafting of config.json would be an easy way to proceed with this. Please go ahead with that idea!

There is indeed an elevated complexity in this config. This got me thinking about how to best transfer this knowledge to new staff members in the future, should the need arise. Some documentation or perhaps even helper scripts would be needed. I will ruminate on this.

scripts/filter
brinkdp: Ah, I see what you mean! I admit that I only tested my edit with the current pop.gen. data file, and thus have scarce evidence to claim that it could be extrapolated to other BEDs we might encounter. Your suggestion looks solid to me.

kwentine:
Great, I will go ahead an implement in one or more separate branches:
-  `skipAddTrack`
- header-preserving sorting of BED files
- tabifying of BED files

When all these are reviewed and merged, I will help you merge back the changes in this branch.

Does that sound good ? The alternative would be for me to implement everything in this very branch,  but I fear it may mix too many concerns.

brinkdp:
Great, sounds like a good plan!

kwentine:
This PR can now be closed, and worked continued on the re-based branch `gwas-test-continued` (#53)

PR #45, opened 2024-09-10
EB-128: Add refNameAlias file links to downloads table

RMCrean:
Hey, 

As discussed over slack, the goal is to add the ability to download the refNameAlias file. 

As each file will be a URL, the current implementation is to just state the URL in the species corresponding `downloads.md` page. I've so far just updated the template file and the Linum tenue `downloads.md` file as you suggested. 

With the current implementation if the parameter is not provided, then the species will not have the option to download the file. 

With regards to UX, I have made it so that clicking on the "download table as a JSON file link", download the file straight away (by using the [download tag](https://www.w3schools.com/tags/att_a_download.asp) ). This can't be done with the alias files as the file is located on an external website, so I changed the labelling to look like this:
Screenshot from 2024-09-10 10-46-30

If you're happy then I think we can go ahead and add the remaining URLs to the other species and then merge the PR.

### DIFF
--- a/hugo/content/species/linum_trigynum/download.md
+++ b/hugo/content/species/linum_trigynum/download.md
@@ -1,4 +1,5 @@
 ---
+alias_file_url: "https://raw.githubusercontent.com/ScilifelabDataCentre/swedgene/main/scripts/data_stewardship/alias_files_temp_storage/CAMGYJ01.fna.alias"
 # The below parameters were generated automatically and do not need to be changed.
 title: "Download"
 layout: "species_download"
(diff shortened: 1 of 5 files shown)


brinkdp:
Thanks, LGTM! Went ahead and merged in order to test the raw file for a bug fix for _L. trigynum_ alias (it now works as intended again). 

Good to know about the download tag. Eventually when I switch to using Figshare to host the aliases, their urls will run a (java?)script on their server side to start an auto-download. So in all, we will likely get the effect anyway 😄

RMCrean:
Awesome, thanks for doing that, and good to know r.e. figshare auto-download.

PR #46, opened 2024-09-25
EB-102: Enable caching images/css/js/fonts on the browser

RMCrean:
This PR is to enable caching images, fonts, js and css files on a user's browser, which is done by editing the nginx config. 

I have tested these changes
**1. locally:** 
```
docker build -t swg-hugo-site:local -f docker/hugo.dockerfile .
docker run -p 8080:8080 swg-hugo-site:local
```
**2. on the [dev cluster](https://swedgene.scilifelab-2-dev.sys.kth.se/)**
using the Hugo docker image created by this branch.

As can be seen from the screenshot the images (and css/fonts etc..) now include a cache policy in their response header so it is behaving as expected. The 2nd load time for a page is now about 1/8 the speed of the initial load time too, so it seems to have had a very nice impact :) 

Screenshot from 2024-09-25 10-05-49

**Notes:** 
- Regarding cache busting: The `styles.css` file created is "[fingerprinted](https://gohugo.io/functions/resources/fingerprint/)" (random string added to the file name on build so cache busting is automatic.  This is not the case for the images though so we'd either need to implement something like that or make sure to use a different image name if we update an image somewhere. 
- For cache lifetimes I just picked some values that seemed to be recommended from googling, happy to change to alternate values if preferred. 
- It seems like there are other optimizations we could do like [enabling gzip compression](https://developer.chrome.com/docs/lighthouse/performance/uses-text-compression). I haven't experimented with that yet, but could do another PR if this one works out and there is time etc... 
- Before merging this PR, the change to the GitHub workflow file (.github/workflows/hugo-container.yml) will be reverted. This was just to make it easy to test the nginx config changes worked not just locally with docker but on the KTH dev cluster (https://swedgene.scilifelab-2-dev.sys.kth.se/). I've marked this as a draft to remind me of this.

### DIFF
--- a/docker/nginx-custom.conf
+++ b/docker/nginx-custom.conf
@@ -15,14 +15,29 @@ server {
     error_page  500 502 503 504  /50x.html;
     location = /50x.html {
     }
+    
+    # security headers
+    add_header X-Frame-Options "SAMEORIGIN"; # don't allow iframing
+    add_header X-Content-Type-Options "nosniff"; # prevent MIME type sniffing
+    add_header X-XSS-Protection "1; mode=block"; # Cross-site scripting (XSS) filter
 
-    # Cache media and font files for 30 days 
+    # cache media and font files for 30 days 
     location ~* \.(webp|svg|jpg|jpeg|png|gif|woff|woff2|ttf|eot|otf)$ {
         add_header Cache-Control "public, max-age=2592000";
     }
 
-    # Cache css and js files for 7 days 
+    # cache css and js files for 7 days 
     location ~* \.(css|js)$ {
         add_header Cache-Control "public, max-age=604800"; 
     }
+
+    # gzip compression 
+    gzip on;
+    gzip_types text/plain text/css application/javascript application/json application/xml;
+    gzip_proxied any;
+    gzip_min_length 1000;
+    gzip_comp_level 6;
+    gzip_vary on;
+    gzip_buffers 16 8k;
 }
+
--- a/docker/nginx-custom.conf
+++ b/docker/nginx-custom.conf
@@ -5,34 +5,24 @@ server {
                 swedgene.scilifelab-2-dev.sys.kth.se
                 genomes.scilifelab.se;
     absolute_redirect off;
+    root /usr/share/nginx/html;
 
     location / {
-        root   /usr/share/nginx/html;
         index  index.html index.htm;
     }
 
     error_page  404              /404.html;
-    error_page   500 502 503 504  /50x.html;
+    error_page  500 502 503 504  /50x.html;
     location = /50x.html {
-        root   /usr/share/nginx/html;
     }
 
-    # Cache media files 
-    location ~* \.(webp|svg|jpg|jpeg|png|gif)$ {
-        root   /usr/share/nginx/html;
-        expires 30d;
-        add_header Cache-Control "public";
+    # Cache media and font files for 30 days 
+    location ~* \.(webp|svg|jpg|jpeg|png|gif|woff|woff2|ttf|eot|otf)$ {
+        add_header Cache-Control "public, max-age=2592000";
     }
-    # Cache font files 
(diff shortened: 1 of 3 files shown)


docker/nginx-custom.conf
```diff
@@ -0,0 +1,38 @@
...
+        index  index.html index.htm;
+    }
+
+    error_page  404              /404.html;
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+
+    # Cache media files
```
jhagberg, line 20: You could combine this since it has the same rules.
like this:
location ~* \.(webp|svg|jpg|jpeg|png|gif|woff|woff2|ttf|eot|otf)$ {

or is there a specific reason?

docker/nginx-custom.conf
```diff
@@ -0,0 +1,38 @@
+server {
+    listen       8080;
+    # same config for localhost, dev and prod environments
+    server_name  localhost 
+                swedgene.scilifelab-2-dev.sys.kth.se
+                genomes.scilifelab.se;
+    absolute_redirect off;
+
```
jhagberg, line 8: Since the root path is the same across multiple locations, you can move the root directive to the server context for better maintainability:
root /usr/share/nginx/html;
This way, all location blocks inherit the root unless overridden.

docker/nginx-custom.conf
```diff
@@ -0,0 +1,38 @@
...
+    location ~* \.(webp|svg|jpg|jpeg|png|gif)$ {
+        root   /usr/share/nginx/html;
+        expires 30d;
+        add_header Cache-Control "public";
+    }
+    # Cache font files 
+    location ~* \.(woff|woff2|ttf|eot|otf)$ {
+        root   /usr/share/nginx/html;
+        expires 30d;
+        add_header Cache-Control "public";
```
jhagberg, line 30: I am a bit unsure about this but it may be a good thing also to specify max-age in the add_header
add_header Cache-Control "public, max-age=2592000";
since

jhagberg (commented):
Just minor things see my inline comments
It can also be a good thing to turn on gzip

 # Gzip compression
    gzip on;
    gzip_types text/plain text/css application/javascript application/json application/xml;
    
Also investigate if you need any security headers or a Content-Security-Policy
  some examples https://gist.github.com/ambroisemaupate/bce4b760405558f358ae

docker/nginx-custom.conf
RMCrean: Fair point, I have combined them now as part of commit 7500e00953d8cfd91204e9ff65c5230533ecb59f

docker/nginx-custom.conf
RMCrean: Thanks, I didn't know that could be done. I have done that now as part of commit 7500e00953d8cfd91204e9ff65c5230533ecb59f

docker/nginx-custom.conf
RMCrean: Thanks. I did some digging and it turns out [the `expires` header is not needed if `max-age` is included](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires) so I've updated the config to use the max-age param instead in commit 7500e00953d8cfd91204e9ff65c5230533ecb59f

RMCrean:
Thanks @jhagberg for the review it was very helpful. I've turned on GZIP now and added some security headers (commit f23d58330d3e2dcb24815811cc08070cc5bc5cf8) based on the link, and they are working as expected on the dev instance. 

I have made a Jira task to implement a [Content Security Policy](https://web.dev/articles/csp), as I will need to do some refactoring of the Hugo code (as have quite a bit of inline styling etc...) before that can work properly.

kwentine (approved):
Thanks, examining your changes was very instructive!

RMCrean:
Thanks Quentin, merging now after updating the hugo container workflow file

jhagberg (commented):
LGTM

PR #47, opened 2024-09-30
EB-141: Use vanilla JS instead of jquery where possible

RMCrean:
Before this PR, jQuery was added as a dependency for all pages and used in some pages as part of some JS scripts throughout the website. I've now removed it to only be used when the Datatables library is loaded (jquery is a dependency) and replaced the other instances with vanilla JS. This therefore helps to improve the sites' load speed by not having to load jquery from a CDN.  

I've also taken the opportunity to move all downloads of scripts/style sheets to one place, ` head.html`, so it is easier to keep track of. Finally I've added [some preconnects](https://developer.chrome.com/docs/lighthouse/performance/uses-rel-preconnect) too to help with the site's loading speeds.

### DIFF
--- a/hugo/layouts/index.html
+++ b/hugo/layouts/index.html
@@ -105,14 +105,14 @@
     const speciesCards = Array.from(document.querySelectorAll('.scilife-species-card'));
 
     function sortAlphabetically(a, b) {
-        const titleA = $(a).find('#science-name').text();
-        const titleB = $(b).find('#science-name').text();
+        const titleA = a.querySelector('#science-name').textContent;
+        const titleB = b.querySelector('#science-name').textContent;
         return titleA.localeCompare(titleB);
     }
 
     function sortRevAlphabetically(a, b) {
-        const titleA = $(a).find('#science-name').text();
-        const titleB = $(b).find('#science-name').text();
+        const titleA = a.querySelector('#science-name').textContent;
+        const titleB = b.querySelector('#science-name').textContent;
         return titleB.localeCompare(titleA);
     }
 
@@ -124,8 +124,8 @@
     }
 
     function sortUpdated(a, b) {
-        const dateStrA = $(a).find('.hidden-date').text();
-        const dateStrB = $(b).find('.hidden-date').text();
+        const dateStrA = a.querySelector('.hidden-date').textContent;
+        const dateStrB = b.querySelector('.hidden-date').textContent;
 
         const dateA = stringToDate(dateStrA);
         const dateB = stringToDate(dateStrB);
--- a/hugo/layouts/index.html
+++ b/hugo/layouts/index.html
@@ -108,7 +108,7 @@
     function getElementText(root, selector) {
         const element = root.querySelector(selector);
         if (!element) {
-            console.debug(`Element not found for selector: ${selector}`);
+            console.warn(`Element not found for selector: ${selector} in root: ${root}`);
             return '';
         }
         return element.textContent;
--- a/hugo/layouts/index.html
+++ b/hugo/layouts/index.html
@@ -104,9 +104,19 @@
 <script>
     const speciesCards = Array.from(document.querySelectorAll('.scilife-species-card'));
 
+    // helper function to handle missing elements by returning empty string
+    function getElementText(root, selector) {
+        const element = root.querySelector(selector);
+        if (!element) {
+            console.debug(`Element not found for selector: ${selector}`);
+            return '';
+        }
+        return element.textContent;
+    }
+
     function sortAlphabetically(a, b) {
-        const titleA = a.querySelector('#science-name').textContent;
-        const titleB = b.querySelector('#science-name').textContent;
+        const titleA = getElementText(a, '#science-name');
+        const titleB = getElementText(b, '#science-name');
         return titleA.localeCompare(titleB);
     }
 
@@ -120,8 +130,8 @@
     }
 
     function sortUpdated(a, b) {
-        const dateStrA = a.querySelector('.hidden-date').textContent;
-        const dateStrB = b.querySelector('.hidden-date').textContent;
+        const dateStrA = getElementText(a, '.hidden-date');
+        const dateStrB = getElementText(b, '.hidden-date');
 
         const dateA = stringToDate(dateStrA);
         const dateB = stringToDate(dateStrB);
@@ -141,9 +151,9 @@
             cardsToAppend = speciesCards.slice();
         } else {
             cardsToAppend = speciesCards.filter(function (card) {
-                const title = card.querySelector('#science-name').textContent.toLowerCase();
-                const subtitle = card.querySelector('#common-name').textContent.toLowerCase();
-                const cardText = card.querySelector('.card-text').textContent.toLowerCase();
+                const title = getElementText(card, '#science-name').toLowerCase();
+                const subtitle = getElementText(card, '#common-name').toLowerCase();
+                const cardText = getElementText(card, '.card-text').toLowerCase();
                 return title.includes(searchText) || subtitle.includes(searchText) || cardText.includes(searchText);
             });
         }
--- a/hugo/layouts/index.html
+++ b/hugo/layouts/index.html
@@ -110,11 +110,7 @@
         return titleA.localeCompare(titleB);
     }
 
-    function sortRevAlphabetically(a, b) {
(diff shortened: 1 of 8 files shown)


hugo/layouts/index.html
```diff
@@ -105,14 +105,14 @@ <h3 class="card-title" id="common-name">{{ .Params.subtitle | markdownify }}</h3
...
 
     function sortAlphabetically(a, b) {
-        const titleA = $(a).find('#science-name').text();
-        const titleB = $(b).find('#science-name').text();
+        const titleA = a.querySelector('#science-name').textContent;
+        const titleB = b.querySelector('#science-name').textContent;
         return titleA.localeCompare(titleB);
     }
 
     function sortRevAlphabetically(a, b) {
```
kwentine, line 113: A common trick to reverse sorting is to negate the original sorting function:
```js
sortRevAlphabetically = a, b => -sortAlphabetically(a, b)
```

hugo/layouts/index.html
```diff
@@ -105,14 +105,14 @@ <h3 class="card-title" id="common-name">{{ .Params.subtitle | markdownify }}</h3
     const speciesCards = Array.from(document.querySelectorAll('.scilife-species-card'));
 
     function sortAlphabetically(a, b) {
```
kwentine, line 107: A nitpick: it could be worth handling the case where `.querySelector` returns `null`. This would currently raise an attribute error. A possible way to go (although certainly not definitive) : define a `getElementText(root, selector)` function that returns the text content of the element under root retrieved by given selector, or the empty string if no such element exists. Or just using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) maybe,

hugo/layouts/partials/head.html
kwentine, line 1: Since the `head.html` partial is used only in `_default/baseof.html` I wonder if it would not be easier to inline the header, and define blocks that could be overridden in individual layouts.

After including common scripts you could provide: 
```
{{- bock "script_includes" . }}{{- end }}
```

And then in `species_intro.html` for example: 
```html
<link rel="preconnect" href="https://unpkg.com">
<link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"
            integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY=" crossorigin="" />
        <script src="https://unpkg.com/leaflet@1.9.4/dist/leaflet.js"
            integrity="sha256-20nQCchB9co0qIjJZRGuk2/Z9VM+kNiyxNV1lvTlZBo=" crossorigin=""></script>
```

This maybe has the advantage of declaring dependencies where they are needed/used in the codebase. And is also more robust if a Squirrel spell checker would decide that "Download" page should be renamed "Downloads".

hugo/layouts/partials/head.html
```diff
@@ -25,8 +25,19 @@
     <!-- Establishing early connections to speed up downloads -->
     <link rel="preconnect" href="https://fonts.googleapis.com">
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-    <link rel="preconnect" href="https://code.jquery.com">
     <link rel="preconnect" href="https://cdn.jsdelivr.net">
```
kwentine, line 28: I wonder: why `preconnect` instead of `preload` ? I read in the docs you linked to: "In general, try to use <link rel="preload">, as it's a more comprehensive performance tweak" 🤔

hugo/layouts/partials/distrib_map.html
```diff
@@ -35,20 +35,12 @@ <h4 class="me-auto">Help</h4>
...
-
+<!-- Leaflet observation map script -->
 {{ $latitude := .Params.latitude | default 0 }}
 {{ $longitude := .Params.longitude | default 0 }}
 {{ $initialZoom := .Params.initialZoom | default 1 }}
 
-
 <script>
-
     const map = L.map('map', {
```
kwentine, line 44: To speed up page load, have you considered adding the `defer` attribute to the leaflet script, and wrapping the map code in an event listener `document.addEventListener("DOMContentLoaded", function() { /* Map code here */ }` ? I'm not sure about this but wanted to evoke the possibility.

hugo/layouts/partials/head.html
```diff
@@ -53,18 +64,30 @@
...
-    <!-- JQuery -->
-    <script src="https://code.jquery.com/jquery-3.7.1.min.js"
-        integrity="sha256-/JqT3SQfawRcv/BIHPThkBvs0OEvtFFmqPF/lYI/Cxo=" crossorigin="anonymous">
-    </script>
+    <!-- Datatables (sometimes needs Jquery as dependancy) -->
+    {{ if or (eq .Title "Glossary") (eq .Title "Download") (eq .Title "Genome assembly") }}
+        <script src="https://code.jquery.com/jquery-3.7.1.min.js"
+            integrity="sha256-/JqT3SQfawRcv/BIHPThkBvs0OEvtFFmqPF/lYI/Cxo=" crossorigin="anonymous">
+        </script>
+        <link href="/Datatables/datatables.min.css" rel="stylesheet">
```
kwentine, line 72: Point of curiosity: is there a reason why we don't rely on a CDN for Datatables ?  Or even JBrowse for that matter. I suddenly realize that I'm perhaps not quite aware of the tradeoffs.

kwentine (approved):
Thanks a lot for this refactoring, and all the research behind it concerning site-load optimization.  

My comments are mostly peripheral (not directly in the scope of the PR) and I see nothing truly blocking so I will approve and let you pick what you think worthwhile (maybe saving the rest for dedicated PRs) 

The most important reflection maybe concerns the usage block to include page-specific dependencies. 

And finally, I wanted to share this [O'Reilly book](https://www.oreilly.com/library/view/high-performance-browser/9781449344757/) that I studied some years ago, and that you might enjoy  🤓

hugo/layouts/index.html
RMCrean: Thanks for the tip, I hadn't thought of that! Applied in commit 171acb4fb80819c65be5362a561d68c08c8c5518

hugo/layouts/partials/head.html
RMCrean: That's really nice! I think this is definitely something we should implement in another PR. I've added it to Jira (EB-196) now :)

hugo/layouts/partials/head.html
RMCrean: Thanks, bit weird that lighthouse makes the warning about preconnect but the recommends preload. My new understanding now is that I should probably use preload for everything except things that are non-blocking. I've made a Jira task (EB-197) for this as well as I think it could take some experimentation.

hugo/layouts/partials/distrib_map.html
RMCrean: I see what you mean and think it could help. I'm not sure it matters but it is important perhaps that the map is "Above the fold content", but I can investigate it (Jira EB-198)! :)

hugo/layouts/partials/head.html
RMCrean: I think this [stackoverflow post](https://stackoverflow.com/questions/42591235/is-it-better-to-use-cdn-for-js-and-similar-resources-vs-local) was quite useful, so it seems like it could indeed be worth using a cdn instead. I've made a Jira task for this to investigate it (EB-200).

kwentine: I have on my side examined a bit more the CDN option for JBrowse, and in that case there might be good reasons to stick with our hosted assets. In short, the entry point exposed by the JBrowse CDN builds is lower level, we get more out of the box (e.g. plugin loading, relative URLs in `config.json`...) by using the jbrowse web releases. But I may start an experimental PR.

hugo/layouts/index.html
kwentine: I think the console message is fine. I would perhaps use `console.warn`, and mention `root` as well in some way or another to help pinpoint the faulty DOM node.

My ideal would be an array of JSON objects that acts as a source of truth, which is rendered every time a sorting is requested. But this is way over-engineered for our purposes.

RMCrean: Thanks for the suggestions, I've just done that now in commit f459afa25720274945bceac9172cebe5afd7c220

PR #48, opened 2024-09-30
Ensure BED files are tab-delimited, sorted, and include headers 

kwentine:
- Ensure BED files are tab-delimited
  - As [discussed](https://scilifelab.atlassian.net/browse/EB-131?focusedCommentId=15732) indiscriminately replace every sequence of spaces by a single tab. Let's see how far that gets us.
- Sort BED files while preserving the header lines

### DIFF
--- a/scripts/filter
+++ b/scripts/filter
@@ -1,12 +1,32 @@
 #!/bin/bash
 #
 # Apply file-type-specific filters to stdin
-#
+
+
+# Sort BED file
+# Output:
+#   Sorted BED stream with headers preserved.
+sort_bed() {
+    local -r tmp=$(mktemp)
+    cat > "${tmp}"
+    headers=(-e '^browser' -e '^track' -e '^#')
+    grep -E "${headers[@]}" "${tmp}"
+    grep -E -v "${headers[@]}" "${tmp}" | sort -k1,1 -k2,2n
+    rm -f "${tmp}"
+}
+
+# Replace spaces by tabs
+# Output:
+#  Input stream with every whitespace sequence replaced by a
+#  single tab character.
+tabify() {
+    :
+}
+
 # Arguments:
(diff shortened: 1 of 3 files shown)


scripts/filter
```diff
@@ -1,12 +1,32 @@
...
+    rm -f "${tmp}"
+}
+
+# Replace spaces by tabs
+# Output:
+#  Input stream with every whitespace sequence replaced by a
+#  single tab character.
+tabify() {
+    sed -E 's/ +/\t/g'
+}
```
RMCrean, line 24: Not sure if this is an edge case that could actually happen or not but if there was both a space and tab then then the original tab would also be kept. Could perhaps be changed to:
`sed -E 's/\s+/\t/g'`


As an example, this input:
```
input="chr1  100  200\n\
chr1  150\t 250\n\
chr2\t100 200"
```

would give: 
```
Original tabify function output:
chr1    100     200
chr1    150             250
chr2    100     200

Changed tabify function output:
chr1    100     200
chr1    150     250
chr2    100     200
```

RMCrean (approved):
I'm not sure how relevant/realistic my comment is. LGTM :)

brinkdp (approved):
Thanks, looks good! Tested with `dockermake` and a few different BED files (with and without headers; space or tab delimited). All resulted in working JBrowse instances.

(I love the branch name, by the way!)

scripts/filter
brinkdp: Good thinking. I would however consider this an edge-case 😄. A handcrafted BED could indeed end up having this typo, but most BEDs we will encounter have been produced by established bioinformatics tools that (I assume) will apply consistent formatting.

kwentine: Thanks for the suggestion, I think it is a generalization that will safely constrain the output. Added in 47e1911

PR #49, opened 2024-10-01
EB-130: Add meta descriptions for each page. 

RMCrean:
Before this PR only the homepage had a meta tag with a description (found in the <head> of each HTML file). [Lighthouse explains nicely why this is good to have on each page for search engine optimisation](https://developer.chrome.com/docs/lighthouse/seo/meta-description/) 

**The homepage has for example this description:**
`<meta name="description" content="A platform facilitating access and discovery of genome data of non-model eukaryotic species studied in Sweden, hosted by SciLifeLab">`

Now each page is assigned one using the Hugo partial `hugo/layouts/partials/create_description.html`. This includes an automated way to make descriptions for new species so we don't have to worry about that either.

### DIFF
--- a/hugo/layouts/partials/create_description.html
+++ b/hugo/layouts/partials/create_description.html
@@ -13,7 +13,7 @@
     {{ $description = "Learn more about the Swedish Reference Genome Portal and its mission." }}
 
 {{ else if eq .Title "Information på svenska" }}
-    {{ $description = "Lär dig mer om den Swedish Reference Genome Portal och sin uppdrag." }}
+    {{ $description = "Lär dig mer om Swedish Reference Genome Portal och dess syfte." }}
 
 {{ else if eq .Title "Privacy Policy" }}
     {{ $description = "The privacy policy applicable to the Swedish Reference Genome Portal." }}
--- a/hugo/layouts/partials/create_description.html
+++ b/hugo/layouts/partials/create_description.html
@@ -1,0 +1,59 @@
+<!-- creates the site meta description for each page, good for search engine optimization -->
+
+{{ $description := "" }}
+
+{{ if .IsHome }}
+    {{ $description = "A platform facilitating access and discovery of genome data of non-model eukaryotic species studied in Sweden, hosted by SciLifeLab" }}
+
+<!-- Individ pages -->
+{{ else if eq .Title "Contact" }}
+    {{ $description = "Contact us for more information and support." }}
+
+{{ else if eq .Title "About" }}
+    {{ $description = "Learn more about the Swedish Reference Genome Portal and its mission." }}
+
+{{ else if eq .Title "Information på svenska" }}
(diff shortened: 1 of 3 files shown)


RMCrean:
@brinkdp, maybe you have already seen the Trivy scan fail elsewhere and have also looked into it. It's not failing due to a security issue etc... [but instead because of rate limiting/too many requests to download a db](https://github.com/aquasecurity/trivy/discussions/7538). So we can safely ignore the failed check, in this case.

hugo/layouts/partials/create_description.html
```diff
@@ -0,0 +1,59 @@
+<!-- creates the site meta description for each page, good for search engine optimization -->
+
+{{ $description := "" }}
+
+{{ if .IsHome }}
+    {{ $description = "A platform facilitating access and discovery of genome data of non-model eukaryotic species studied in Sweden, hosted by SciLifeLab" }}
+
+<!-- Individ pages -->
```
brinkdp, line 8: Would it be useful for the SEO if each individual page description mentions "The Swedish Reference Portal" by full name? Many of them already do, but not all. Or is it enough if only Home does that? (Perhaps it would be a form of keyword spamming?)

hugo/layouts/partials/create_description.html
```diff
@@ -0,0 +1,59 @@
...
+{{ else if eq .Title "Contact" }}
+    {{ $description = "Contact us for more information and support." }}
+
+{{ else if eq .Title "About" }}
+    {{ $description = "Learn more about the Swedish Reference Genome Portal and its mission." }}
+
+{{ else if eq .Title "Information på svenska" }}
+    {{ $description = "Lär dig mer om den Swedish Reference Genome Portal och sin uppdrag." }}
```
brinkdp, line 16: Small suggestion to make the Swedish sound a tad bit more formal:

Lär dig mer om Swedish Reference Genome Portal och dess syfte.

brinkdp (approved):
Checked-out and tested with the Firefox web dev tools ("F12"). Looks nice, thank you! I had two very minor comments, but other than that, please go ahead and merge.

Yes, I noticed on Friday that Trivy failed me from time to time. I came to the same conclusion as you, that it was a matter on their side and not ours. I'd hoped that it would be resolved by now, but lets see how this develops.

hugo/layouts/partials/create_description.html
RMCrean: I'm honestly not sure, it could be that we are over- or under-using "Swedish Reference Genome Portal". The examples I've seen on the internet don't often seem to mention the website name/title in the description so I suspect it is not necessary/essential at least.

hugo/layouts/partials/create_description.html
RMCrean: I'm glad you checked that one, changed it with commit ba9a44d41fd262d707af4c031b6573c9d68f9507

RMCrean:
@brinkdp, thanks for going through and checking, merging now :)

PR #50, opened 2024-10-02
EB-100. Add Krill species

brinkdp:
This PR regards the addition of data and content for the Krill to the Genome Portal. Most of this is standard procedure for new species entries, but the new challenge here is the substantial size (19 Gb) of the assembly. 

I would specifically like to hear your thoughts on whether this poses any technical challenges. The main thing I have noticed is that it takes >20 min to generate the index for the assembly. But that should hopefully be a one-time operation.

From a user experience point-of-view, the assembly poses other challenges as well. Navigating the assembly is cumbersome (to put it mildly), since there are >200,000 scaffolds. But since this is directly due to the low contiguity of the data, I think that is aspect is outside of our control.


RMCrean (approved):
It looks great, I have some minor suggestions on the webpages. 

I'm still downloading/running the data builder job locally so I can add some comments on that later but I think my response will be that we should merge, as I can't see how we will run into any issues with this data included. And, I agree the UX issue is unfortunate for this species but that is the nature of the data.

brinkdp:
Thanks! Great that you spotted the outdated parameters and some awkward phrasings in the text.

RMCrean:
No problem, I can confirm the data builder job worked for me with the krill included and shows up nicely on the browser. Long way of saying LGTM. :)

config/meganyctiphanes_norvegica/config.yml
```diff
@@ -0,0 +1,13 @@
+organism: "Meganyctiphanes norvegica"
+assembly:
+  name: Mnor_HA_v2
+  displayName: "Meganyctiphanes norvegica genome assembly"
+  accession: GCA_964058975.1
+  # The following url points to the original ENA upload of the assembly. It will require an alias file to be able to load the protein coding genes track
+  url: "ftp://ftp.ebi.ac.uk/pub/databases/ena/wgs/public/cax/CAXKWB01.fasta.gz"
+  # NOTE: this alias file is 15 Mb. Copied it to local hugo/public/data dir for now to enable local testing
+  aliases: "https://raw.githubusercontent.com/ScilifelabDataCentre/swedgene/main/scripts/data_stewardship/alias_files_temp_storage/CAXKWB01.fna.alias"
```
kwentine, line 9: I think we should avoid to put large file into the `main` branch. Something I realized not so long ago was that once you commit a voluminous file, it stays in the `git` database forever, even if it is deleted at a later time from the working tree. So all future `git clone` operations (including maybe some of them run during CI) could potentially be slowed down.

An alternative could be to use a public Nextcloud or Google Drive folder to host temporary voluminous files.

kwentine (approved):
Great, I'm ~~krilled~~ thrilled to see this addition to our bestiary. My CPU is still crunching at the files, giving me ample time to pause and wonder: to what extent can`tabix` and `sort` be parallelized ? I will make a note to investigate.

config/meganyctiphanes_norvegica/config.yml
brinkdp: Agreed! I have also had my regrets about adding this file to the repo. It's interesting, but also a little frustrating, to learn it can affect future performance even if deleted. Well, lesson learned. Another case where I have noticed that we encounter large files is the photos for the species. For cases where the original resolution is high, the files can be considered large for git even after .webp conversion. I forget what the default size limit for committing a file is, but from what you are saying it makes sense to me to not override it from now on.

The alias files will soon go on Figshare for persistent sharing, but for testing I agree that we should consider other hosts. I now also realize that the current implementation is based on reading the alias remotely. Would it be useful if the makefile downloaded remote alias files to disk/ volume just like the assembly and tracks? And if so, would we be interesting in supporting compressed alias files to speed up the downloads?

brinkdp:
> My CPU is still crunching at the files, giving me ample time to pause and wonder: to what extent can`tabix` and `sort` be parallelized ? I will make a note to investigate.

If I recall correctly, `samtools faidx` supports multi-threading. But perhaps, as you say, it might be the other unit operations in the processing step that pose the bottleneck.

> we had humbly [asked](https://github.com/ScilifelabDataCentre/kube-swedgene/blob/f702ef9fa7555711d8a2388376edb72a6fb3189a/base/pvc.yaml#L12) Kubernetes for 10Gi of storage.

Ah, yes, that would be important to contemplate 😄. I think we got indications that we could expect more species in the future that would have a size in the Gb order of magnitude.

config/meganyctiphanes_norvegica/config.yml
RMCrean: > Another case where I have noticed that we encounter large files is the photos for the species. For cases where the original resolution is high, the files can be considered large for git even after .webp conversion. 

Sorry to hijack this thread, you can change the compression level with the cwebp command when doing the conversion if you want to reduce the file size. [see the confluence add new species doc](https://scilifelab.atlassian.net/wiki/spaces/DSNEB/pages/2858352695/Creating+a+new+species+entry). I just had look and the images are less than 500 KB each so not too big but could be worth testing reducing them as faster when a user visits the site.

brinkdp: > you can change the compression level with the cwebp command when doing the conversion if you want to reduce the file size. 

Thanks, agreed! This is indeed what I did to get the files down below that size 😄. I recall that there were cases where I had to iterate on this a few times before finding a compression level that gave a file size below the git limit. Anyway, lesson learned to actively respect the limit and not be tempted to override it 😉.

kwentine: > I have also had my regrets about adding this file to the repo. It's interesting, but also a little frustrating, to learn it can affect future performance even if deleted.

Have no regrets, it was a clever way forward and the file is certainly not huge! The impact is also to be nuanced: sure, developers needing the whole history (3 on the planet right now) will bear the cost once, but it is also possible to use shallow `git clone` strategies if push comes to shove.

> I now also realize that the current implementation is based on reading the alias remotely. Would it be useful if the makefile downloaded remote alias files to disk/ volume just like the assembly and tracks?

In a sense, alias files will always be read remotely since they are queried at runtime by JBrowse, running in the user's browser. So I'm not sure it would be advantageous to store them on our servers. Unless perhaps we want to systematically compress third party aliases. But we would need to study the performance improvement to justify it (and verify that JBrowse supports Gzipped aliases 🤔 )

brinkdp: > In a sense, alias files will always be read remotely since they are queried at runtime by JBrowse, running in the user's browser. So I'm not sure it would be advantageous to store them on our servers.

I see, that makes sense. Let's keep an eye on how the Krill session performs once we start calling the alias from Figshare instead of the GitHub repo; hopefully there will be no noticeable difference.

PR #51, opened 2024-10-03
Support tracks manually added to JBRowse configuration

kwentine:
Elements of the `tracks` list in `config.yaml` can now include an optional `addTrack` property. If omitted, the behavior is unchanged. When set to `false`, the data for the track will be prepared, but it will not be added to `config.json` using `jbrowse add-track`.  Intended use case: tracks already configured manually (along with the default session) in `config.json`.


brinkdp (approved):
Thanks! Works well and adds useful flexibility for species configurations.

RMCrean (approved):
LGTM :)

PR #52, opened 2024-10-03
EB-179: Configure production and development Docker image builds

RMCrean:
Hey :) 

We have 5 workflows, 2 of which build images and the other 3 are more testing related. 

For the 3 testing related I have set it up so that both a PR or push to main or prod triggers them to run. 

**For the 2 that build images I have now set it up so that:** 
- On PR no images are built.
- On push to prod, both images are built with tag "prod".
- On push to main both images are built with tag "dev". 
- A workflow_dispatch (manual job) can be triggered to build either image on any branch (but not a fork). The default tag for the workflow dispatch will be "dev", but we have the choice to change that if we want when triggering the workflow. 

I think this matches up with our plan. I've been testing out these settings using a personal repository and there it seems to be working nicely at least. 

Related to this is of course the configuration of the branch settings, you can see those I have setup here: https://github.com/ScilifelabDataCentre/swedgene/settings/branches 

Cheers!

### DIFF
--- a/.github/workflows/data-builder-container.yml
+++ b/.github/workflows/data-builder-container.yml
@@ -1,15 +1,16 @@
 name: Build and push the data builder image to GitHub Container Registry
 on:
   push:
-    tags:
-      - 'v*.*.*' # i.e. has to have a release tag to run 
+    branches:
+      - prod # aka prod/stable
+      - main # aka dev
       
-  # if run as workflow_dispatch, these tags can be used in build+push step
   workflow_dispatch:
     inputs:
       docker_tag:
-        description: 'Docker tag for the image'
+        description: 'Docker tag for the image, default is dev'
         required: false
+        default: 'dev'
 
 jobs:
   push-to-container-registry:
@@ -32,6 +33,17 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set Docker tag
+        id: set-tag
+        run: |
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            echo "tag=${{ github.event.inputs.docker_tag }}" >> $GITHUB_OUTPUT;
+          elif [ "${{ github.ref }}" == "refs/heads/prod" ]; then
+            echo "tag=prod" >> $GITHUB_OUTPUT;
+          elif [ "${{ github.ref }}" == "refs/heads/main" ]; then
+            echo "tag=dev" >> $GITHUB_OUTPUT;
+          fi
+
       - name: Create Docker metadata
         id: meta
         uses: docker/metadata-action@v5
@@ -47,9 +59,8 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v5
         with:
-
           push: ${{ github.event_name != 'pull_request' }}
           context: .
           file: ./docker/data.dockerfile
-          tags: ${{ github.event.inputs.docker_tag || steps.meta.outputs.tags }}
+          tags: ghcr.io/scilifelabdatacentre/swg-data-builder:${{ steps.set-tag.outputs.tag }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/data-builder-container.yml
+++ b/.github/workflows/data-builder-container.yml
@@ -2,9 +2,10 @@ name: Build and push the data builder image to GitHub Container Registry
 on:
   push:
     branches:
-      - prod # aka prod/stable
-      - main # aka dev
-      
+      - prod # Branch to be deployed in production, through builds tagged prod
+      - main # Branch to be deployed in development, through builds tagged dev
+  
+  # workflow_dispatch can be used on feature branch to test the site on dev instance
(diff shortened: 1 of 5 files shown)


RMCrean:
P.S. I was looking into the creating new releases/tagging question. I think we could explore this [GH action to make draft releases](https://github.com/marketplace/actions/release-drafter) after each merge to prod. It looks very complete and is actively maintained.

.github/workflows/data-builder-container.yml
```diff
@@ -1,15 +1,16 @@
 name: Build and push the data builder image to GitHub Container Registry
 on:
   push:
-    tags:
-      - 'v*.*.*' # i.e. has to have a release tag to run 
+    branches:
+      - prod # aka prod/stable
```
kwentine, line 5: I would explicit those comments, because even as an insider it bends my neurons. 
Suggestions:
-  Branch to be deployed in production, through buillds tagged `prod`
-  Branch to be deployed in development, through builds tagged `dev`

.github/workflows/data-builder-container.yml
```diff
@@ -1,15 +1,16 @@
 name: Build and push the data builder image to GitHub Container Registry
 on:
   push:
```
kwentine, line 3: Since we opened the possibility to dispatch builds manually, I wonder if we could limit automated `dev` builds to when feature branches are merged into `main`. My only concern, that is not solidly grounded, is that we might trigger many builds by pushing small fixes to `main` that do not warrant re-builds (e.g. `Update README`)

kwentine: And on the related note of avoiding unnecessary builds I make this note for later inquiry: if we make a documentation PR to main, is there a way to `mark` the particular PR so that merging it do not trigger a new image build ?

The most general version of this question would be: could we skip re-builds for PR that touches only a set of files (for example always avoid rebuilds of hugo site when only the `scripts` directory receives changes.

This is all food for thought, should we happen to have a light sprint or something.

.github/workflows/data-builder-container.yml
```diff
@@ -47,9 +59,8 @@ jobs:
       - name: Build and push
         uses: docker/build-push-action@v5
         with:
-
           push: ${{ github.event_name != 'pull_request' }}
           context: .
           file: ./docker/data.dockerfile
-          tags: ${{ github.event.inputs.docker_tag || steps.meta.outputs.tags }}
+          tags: ghcr.io/scilifelabdatacentre/swg-data-builder:${{ steps.set-tag.outputs.tag }}
```
kwentine, line 65: Neat! Is there a reason why the complete image name is now needed ?

.github/workflows/hugo-container.yml
kwentine, line 1: Seeing the similarity between the hugo container and data builder workflow files makes me realize that these images are built under the same conditions. Do you see a way of avoiding this duplication ?

kwentine (approved):
Thanks, as always very instructive to review, and conjuring up ideas that are not blocking but maybe worth pondering on :)

.github/workflows/data-builder-container.yml
RMCrean: Fair point, I've now added this to each workflow file to be more clear in commit 807fa1b.

.github/workflows/hugo-container.yml
RMCrean: The way I know for this is using a matrix strategy (see e.g. here https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow), it certainly does reduce duplication but I do think it makes it a bit harder to read. Happy to consider changing though in another PR if needed. I know [serve uses this approach to build there images](https://github.com/ScilifelabDataCentre/serve/blob/develop/.github/workflows/publish.yaml)

.github/workflows/data-builder-container.yml
RMCrean: Great idea, [this is quite straightforward to setup which is nice](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), and I have something similar in the lighthouse.yml gh action already. 

I've now added a setting to prevent rebuilds if only the README.md file is updated in commit 3cc2e149c64c227e84d2b972ef30c3bd7c9a22ee. This could be made better after a good think about exactly what files to commit but I think this should suffice for now.

.github/workflows/data-builder-container.yml
RMCrean: It's a bit weird that it seems to be needed but [when I tested on my own testing repo](https://github.com/RMCrean/multi-docker-action/commit/40ef254274076bfb678a5a4a57ac9d96c38574ea) (to prevent polluting this repo with failed builds/commits) it broke the build without it. 

I think this is because the docker/metadata-action@v5 we were using before handled this automatically, whereas with the custom bit of logic in Set Docker tag there is no automatic handling.

RMCrean:
Thanks @kwentine I will merge now even if we have some non-blocking comments/ideas still going so I can finish off work on the rename of the repo. :)

PR #53, opened 2024-10-09
GWAS plugin test (continued)

kwentine:
This draft PR is a continuation of #43, now that the identified [prerequisites](https://github.com/ScilifelabDataCentre/swedgene/pull/43#issuecomment-2363298675) have been integrated into the `main` branch.

I have cherry-picked the commits that seemed most relevant, and replaced the former placeholder track by a manually-added one (using `addTrack: false`).


brinkdp:
Thanks for this! Added [7f1fd55](https://github.com/ScilifelabDataCentre/genome-portal/pull/53/commits/7f1fd55f34609d811238807eec7e5cc868dbc74e) for the sake of posterity, should we ever need to revisit the log from this PR. It irons out some bugs from test data.

I took a somewhat lazy approach to get the `gwas_testing_ground` dummy data to display. Since the dummy has a .txt extension and as such will not be parsed by the makefile, I renamed the extension to .bed via `fileName` in `config.yml`. I was amused to see that it worked 😏 It's a fun hack for testing purposes, but is probably a practice that we should avoid for real data.

I have also figured out the cause for the behavior reported in the note in [#43](https://github.com/ScilifelabDataCentre/genome-portal/pull/43#issuecomment-2331756845). It was the the sparsity of the data (100kbp windows) that made it look like it was not loaded properly when `defaultSession` was configured to higher zoom levels. Increasing `bpPerPx` solved this.

Looks good to me!

kwentine:
I will then delete the old `gwas-test-data` branch. Is this PR intended to be included in the first release ? If you intend to keep it as an experiment I think we can close the PR (but keep the branch if you want to use it as a starting point for further tests)

> I renamed the extension to .bed via `fileName`

Nice! The `Makefile` takes all the downloaded files as a starting point for further processing, and `fileName` tells how local copy should be named. So it works as expected (although `fileName` was first intended to be used with opaque URLs)

brinkdp:
Sounds good! I am happy about the findings of this experiment for now and I think we can merge this and include it in the first release.

PR #54, opened 2024-10-10
EB-124: Update GitHub Repo Name

RMCrean:
Repo renamed from swedgene to genome-portal and this PR is to update all instances of the word. The PR is marked as a draft as I'm not sure on something. 

The single commit I've made so far is just to update the URL of the repo when used, so relatively harmless. There are still **11 instances** of `swedgene` in the repo. 

The instance in `docker/nginx-custom.conf` should not be changed as this is the url of the dev site. The remaining 10 I'm less sure about:

For example, in `scripts/dockermake` the following line is used: 
```docker_workdir='/swedgene'```

If I understand correctly, the current form in this branch would still work for us whilst our Git repo is still locally called `swedgene` but for a new dev or if one of us were to rename the git repo folder locally to `genome-portal` this would break? That said I tried this locally (renaming my repo to `genome-portal`) and remaking the images locally on this branch and running `make` worked, so perhaps it is fine?  @kwentine you can maybe shed some light here perhaps?

If it is a problem then I think we should each rename our local git repositories from "swedgene" to "genome-portal" and we can update these instances as perhaps that is best long term (if we get a new dev/computer etc...)? 

**Happy for one of you to add commits to this PR if you feel comfortable with making the needed changes!**


kwentine:
Thanks for the initiative!   

> For example, in `scripts/dockermake` the following line is used: `docker_workdir='/swedgene'`

This particular occurrence refers to the working directory in Docker image (`WORKDIR /swedgene`), and hence not impacted by changing the repo name. Changing it though requires care, since `kube-swedgene` manifests rely on this to define mount points.

I will have a pass at the remaining occurrences and post a message when I am done committing.

kwentine (approved):
All good on my side. I created a [JIRA issue](https://scilifelab.atlassian.net/browse/EB-219) to handle `swedgene` occurences in Dockerfiles, since those impact our Kubernetes manifests.

RMCrean:
Awesome, thanks I will merge then :)

PR #56, opened 2024-10-17
EB-188: Ensure README is presentable for release

RMCrean:
Hey @kwentine , this is as far as I got with the README, thanks for taking over! 

Part of me thinks we could split the development info into a separate file but I'll leave you to decide on that. 
Let me know when you want me to take a look or if you want me to write about something specific! 

Cheers


RMCrean:
Hey, I think the README looks much better thanks. 

I think we can remove the section on branches and fix some of the table of contents links, otherwise I think it looks good. We can have a separate markdown document on adding a new species (as it already exists on confluence) and that can cover things like pre-commit etc.. anyway.

PR #59, opened 2024-10-22
EB-218: Add testing to the website using playwright

RMCrean:
This PR is to add [Playwright](https://playwright.dev/python/) tests for the website. 

- The file `playwright/Running_Playwright.md` contains some documentation on how to run the tests. 

One thing I'm pleased about with these tests is that new species pages will be automatically picked up by the test suite. So this should be a good way to catch issues when a new species is added.   

Timing wise, when run in parallel it takes about 45 seconds on my laptop (about ~15 seconds of which is waiting 3 x 5 seconds for each failed assertion test to Timeout - default is 5 seconds). 

Additional Notes:

- **This PR does not include a GH actions workflow to run the tests**. I think this should be a separate PR later instead. This is because some tests (3) are currently failing (but that is because they should). I therefore want to fix those (either by changing the test or changing something on the site) with Daniel first. 

- There is an async form of playwright but [from some googling it does not seem like a good idea to use](https://www.reddit.com/r/softwaretesting/comments/198s7ap/python_playwright_with_async_methods/).


brinkdp (approved):
Nicely done, thanks for implementing this! 🙌 

I got an error when following the README, but it was self-explanatory: 

```
E           ║ Looks like Playwright was just installed or updated.       ║
E           ║ Please run the following command to download new browsers: ║
E           ║                                                            ║
E           ║     playwright install  
```
Maybe it's worth adding a comment to the doc about running `playwright install` when installing it for the first time? 

The tests themselves look very useful and having a GH action sounds like a good idea. Would be run for each commit, or at a later stage (say, at PR merge or deployment)?

Good that the automation spotted the missing _L. trigynum_ heading. I had forgotten about that! I had deliberately removed it since there were no other references than the genome reference, which was already stated under another heading (listing the same reference twice in a row just looked silly). I plan to fix this in another branch when we implement the new version for how to cite a species entry and its data.

playwright/conftest.py
```diff
@@ -0,0 +1,111 @@
+"""
+Essentially holds the fixtures for each page on the website.
+
+Note that the use of context.new_page() is required when you combine multiple page objects in a single test.
+For example when generating the all_pages() fixture, we need to create a new page object for each page.
```
kwentine, line 5: I have wondered quite a lot about this 🤔  Would tests  parameterized with lists of strings, and using the function-scoped fixture `page` work ?
```python
@pytest.mark.parameterize("url", ['/foo', '/bar'])
def test_page(url, page):
  page.goto(url)
  # etc...
```
I'll push a litter helper to generate all Hugo URLs

```

**Review by kwentine — COMMENTED** [review]
This is a first review after discovering Playwright. Thanks for educating us about such an interesting tool!

File: playwright/Running_Playwright.md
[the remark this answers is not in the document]
Comment (RMCrean) [reply playwright/Running_Playwright.md:1]: Yes, but I personally prefer the separation because it's extra obvious what files are associated with what etc.. I don't have a strong preference here though so happy to move if you prefer?

File: playwright/conftest.py (hunk 1 above)
Comment (RMCrean) [reply playwright/conftest.py:5]: Thanks for the suggestions both here and on slack, I'm going to look into this now on this PR!

**Comment by RMCrean** [comment]
@brinkdp, thanks for the review! 

> Maybe it's worth adding a comment to the doc about running playwright install when installing it for the first time?

Thanks for spotting this, I've added that now to the document, commit XXXX

> The tests themselves look very useful and having a GH action sounds like a good idea. Would it be run for each commit, or at a later stage (say, at PR merge or deployment)?

I think on every commit might be overkill - especially if someone wants to commit WIP. So I think on a PR (to merge into main) would be lagom. If you have a preference here feel free to say though. 

> Good that the automation spotted the missing L. trigynum heading. I had forgotten about that! I had deliberately removed it since there were no other references than the genome reference, which was already stated under another heading (listing the same reference twice in a row just looked silly). I plan to fix this in another branch when we implement the new version for how to cite a species entry and its data.

Glad to hear. If there are other things you think would be nice to test for or if there are some things that we're testing for that are actually too strict (e.g. maybe it is okay to not have this heading, and a warning might be better) then happy to work together on some changes.

File: playwright/Running_Playwright.md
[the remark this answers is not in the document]
Comment (kwentine) [reply playwright/Running_Playwright.md:1]: No I don't mind, but it's always good to have a random comment to get a review started 😅

**Comment by brinkdp** [comment]
> I think on every commit might be overkill - especially if someone wants to commit WIP. So I think on a PR (to merge into main) would be lagom. If you have a preference here feel free to say though.

Like the saying goes: "Lagom är bäst". I _am_ fond of WIP commits and our current use of placeholders, so this sounds good to me!

> Glad to hear. If there are other things you think would be nice to test for or if there are some things that we're testing for that are actually too strict (e.g. maybe it is okay to not have this heading, and a warning might be better) then happy to work together on some changes.

Cool, I think it would be a fun exercise to brainstorm together on potential new test cases. Having warnings instead of fails sounds interesting too.

One idea I've had in the back of my head for a while is to write a test that checks that all occurrences of the scientific species name on a page are italicized. I always try to do an extra manual check on this, but it nevertheless slips by now and then. 

A common cause of this is copying texts from researchers from a mail or .docx and forgetting to adjust the syntax for the markdown files. Angela has suggested automating this step with pandoc (which I think is a good idea), but a playwright test could be useful too, if it's not too much work to implement.

In my mind, catching this with pattern matching should hopefully be fairly straight forward since we already declare the scientific name in the `science_name` variable in each species' `_index.md`. We would need to include both the full and abbreviated form: _Tiny herb_ and _T. herb_, and the different syntax options for italics in markdown: `*single asterisk*`, and `_underscore_`. Do you think this is feasible?

File: playwright/conftest.py (hunk 1 above)
Comment (RMCrean) [reply playwright/conftest.py]: The suggestion in slack to use `@pytest.mark.parametrize` with  `indirect=True` worked really nicely. I've now committed those changes in commit [868c5a6](https://github.com/ScilifelabDataCentre/genome-portal/pull/59/commits/868c5a6407ec0988c13d50fce5e023a7d6b727cc) 

For the helper function to generate the urls, I decided to go for a different approach (so we can use this in a GH actions test). The non-species pages are still hard-coded (as I think these are unlikely to update past launch). To search for the species pages I have kept the same approach of looking up the names of the content folders inside the Hugo site and added a check to make sure the page is not in draft mode (so not yet visible on the site). This is covered in commit [4725094](https://github.com/ScilifelabDataCentre/genome-portal/pull/59/commits/472509473dffd913ba3060980f010288daa07572) A "good enough" solution I think.

**Comment by RMCrean** [comment]
@kwentine, Thanks again for suggestions and ideas. I've made the changes we talked about so it's ready for a 2nd review when you have time. 

@brinkdp, 

> Cool, I think it would be a fun exercise to brainstorm together on potential new test cases. Having warnings instead of fails sounds interesting too.

Definitely up for that! 

> One idea I've had in the back of my head for a while is to write a test that checks that all occurrences of the scientific species name on a page are italicized. I always try to do an extra manual check on this, but it nevertheless slips by now and then.
A common cause of this is copying texts from researchers from a mail or .docx and forgetting to adjust the syntax for the markdown files. Angela has suggested automating this step with pandoc (which I think is a good idea), but a playwright test could be useful too, if it's not too much work to implement.
In my mind, catching this with pattern matching should hopefully be fairly straight forward since we already declare the scientific name in the science_name variable in each species' _index.md. We would need to include both the full and abbreviated form: Tiny herb and T. herb, and the different syntax options for italics in markdown: *single asterisk*, and _underscore_. Do you think this is feasible?

Thanks for the idea, I think this should be feasible, in the case of Playwright we would be checking the species names in the markdown texts has an `<em>` tag as these tests run on the built website not on the markdown files themselves. I will look into it and get back to you :)

**Comment by RMCrean** [comment]
@brinkdp italics checked now added :) I ended up using the markdown files to perform the checks and not the website (i.e. didn't use playwright but just pytest). 

@kwentine, I've updated the tests to reflect the latest changes on main. This is now ready for a review/merge I think.

File: playwright/tests/utils.py [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+        all_non_species_pages_paths() + all_intro_page_paths() + all_assembly_page_paths() + all_download_page_paths()
+    )
+
+
+def validate_date_format(date: str, date_format: str) -> None:
+    """
+    Validate the format of a date on the website is as expected.
+    Returns nothing if fine, will raise Error otherwise, causing test fail.
+    """
+    ALLOWED_DATE_FORMATS = ["%d %B %Y", "%d/%m/%Y"]
```
kwentine, line 103: This variable could be defined at module level. If 1000 dates are being tested (think of the Wallenberg-funded insect species looming over us), as many identical list allocations will occur.

playwright/tests/utils.py
```diff
@@ -0,0 +1,111 @@
...
+            if "draft: true" not in content:
+                species.append(folder.name)
+
+    return species
+
+
+SPECIES_LIST = get_list_of_species()
+
+
+def all_intro_page_paths() -> list[str]:
```
kwentine, line 45: A general question: why did you prefer functions that return constant values rather than module-level constants ? For example, `get_hugo_species_dir()` rather than `HUGO_SPECIES_DIR = ...`

playwright/tests/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    species = []
+    for folder in HUGO_SPECIES_DIR.iterdir():
+        if not folder.is_dir():
+            continue
+
+        index_file = folder / "_index.md"
+        if not index_file.exists():
+            continue
+
+        with index_file.open() as file_in:
```
kwentine, line 34: So here comes a nitpick, but it is a nice thing to ponder about. This code checking for draft status is difficult to test because: 
1. It is nested in a loop
2. It performs IO

Here is a refactoring suggestion:
```python
# Isolate the logic in a function that takes a string as input, because it is easy to test
def is_draft(content):
  return "draft: true" in content

def list_species():
  return [ folder.name for folder in HUGO_SPECIES_DIR.iterdir()  if (index := folder / "_index.md").exists() and is_draft(index.read_text())]
```
- `read_text()` is nice because it handles opening and closing the file
- the walrus `:=` and list comprehension are just ~to show off~ for the curiosity

kwentine (approved):
I am more than happy and impressed by the work, but as usual some philosophical remarks, just hoping they may be interesting.

playwright/tests/utils.py
RMCrean: Thanks for this, it never occurred to me that you could just import the constant into other files *facepalm*. I've done that now in commit 01c94dfccc410fbafdf51d849cb6d28f95fed5fa

playwright/tests/utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def all_download_page_paths() -> list[str]:
+    """
+    Return all species download page paths as strings.
+    """
+    download_paths = []
+    for species in SPECIES_LIST:
+        download_paths.append(f"{species}/download")
+    return download_paths
```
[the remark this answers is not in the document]
RMCrean: Included in commit 01c94dfccc410fbafdf51d849cb6d28f95fed5fa. The utils module is a lot smaller now, thanks!

playwright/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+    Generate a single page object.
+    Helper function to generate unique pages like about, contact, etc.
+    """
+    page = context.new_page()
+    page.goto(url)
+    return page
+
+
+@pytest.fixture
+def home_page(context: BrowserContext, base_url: str) -> Page:
```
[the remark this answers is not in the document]
RMCrean: Thanks, the need to create a new context was due to how I had previously solved the issue of looking at multiple pages in one test (before moving to the much more optimal way of providing a list of pages to test with pytest.mark.parameterize)

I was able to reduce the code to look more like this now :) 
```
def generate_single_page(page: Page, url: str) -> Page:
    page.goto(url)
    return page

@pytest.fixture
def home_page(page: Page, base_url: str) -> Page:
    return generate_single_page(page=page, url=f"{base_url}/")
``` 

As the page.goto() method returns a response object and not the page object, it still has to be 2 lines, which is why I kept the helper function! 

Updated in commit 256fa0ea49df90142ee9db2156d81241987031d0

RMCrean:
Thanks @kwentine, merging now :) I'll soon start work on a making a PR to create a GH actions test for this one too :)

PR #61, opened 2024-10-22
Install JBrowse at build time

kwentine:
What started as a simple endeavor to upgrade our JBrowse bundle ended up in this shift of strategy: instead of having the JBrowse web static bundles under source control (and incurring huge diffs and big file additions every time we upgrade), why not fetch a pinned JBrowse version at build-time ?

This PR:
- Removes `hugo/static/browser` from our source tree 
- Adds an extra step to `dockerfiles/hugo.dockerfile` that fetches the JBrowse assets of interest to us, to be dropped in the NGINX `html` folder.
- Provides a helper `script/download_jbrowse <version> <out_dir>`to fetch and unpack a specified JBrowse release

The JBrowse bundle entrypoint is named with a hash that changes every release: `main.sd34sdf.js`. To smooth this out, I took the approach of renaming it to `main.js`. While this gives a uniform name to refer to, it may be unsatisfactory for cache busting purposes.


RMCrean:
Thanks for this! I like the approach

When I follow the README steps to build the images, install data etc.. on this branch I get a 404 error for the file: `http://localhost:8080/browser/static/js/main.js` in the browser. When I exec'd into the container I found the file `genomeView.js` in that folder which I think comes from `hugo/static/js/genomeView.js `. Perhaps that file should be removed too? 

Is there something I need to differently to get it to work on my end in terms of the protocol do you know?

kwentine:
> When I follow the README steps to build the images, install data etc.. on this branch I get a 404 error for the file: `http://localhost:8080/browser/static/js/main.js` 

My first impulse was to scour the `Chrome` codebase for a bug, but then I thought the error may in fact come from the `README`. So could you try again with updated instructions ? 🙃 

>  I found the file `genomeView.js` 

Got nostalgic and then removed it.

RMCrean (approved):
This is fantastic, it works so smoothly now. 

Thanks for the hard work on this!

PR #62, opened 2024-10-25
EB-223: Swap to trunk based dev + add version number to website footer 

RMCrean:
Hej! I hope the information overload proves useful! 

This PR covers: 

1. The swap to trunk based development in the GH actions. 
2. Making the genome-portal and JBrowse versions show up in the footer of each page.
3. Bonus: I used the matrix_strategy to have just 1 workflow file for building both images :). 

**For part 1,** this essentially can be summarized as removing references to the "prod" branch in the GH actions files. I also took the liberty to remove the lighthouse GH action as find myself never using the action output (and I guess you aren't either?) and instead find running it inside the chrome browser works better. 

**For part 2:** 
- This works by the `hugo.dockerfile` passing a string containing the version info to the `hugo build `command. 
- This string is then interpreted by the Hugo Partial: `hugo/layouts/partials/version_info.html`. The string includes both info about the genome-portal version and the JBrowse version (more on that later). It is the job of the partial to correctly display different build types in the footer in a nice way (e.g. how a release should look vs a build on a feature branch etc...). 

GH actions runs:
- When a GH actions run is triggered, it can also build this string accordingly based on things like the branch it is run on and if it was triggered by a `release`, `push` or `workflow_dispatch`. See below for the different cases it is designed to cover. 

Local development:
- The parameter can be optionally provided when building the Hugo docker image locally with `./scripts/dockerbuild hugo`, (I've tried to follow you're approach here for adding a parameter to that file - please feel free to edit if needed).
- One can test locally how different possible strings will look by passing the `SWG_HUGO_ENV_ARG` flag to the `./scripts/dockerbuild hugo` command - see cases below.

**The cases:** 

**1. A release is published.** This will trigger a build that will set the string to have the form: "release,v0.1.2"
What this will look like can be tested locally by: 

```SWG_HUGO_ENV_ARG="release,v0.1.2" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve```

Screenshot from 2024-10-25 11-02-50

**2. A push to the main branch** - can test locally with: 

```SWG_HUGO_ENV_ARG="development,fbabebb3dfb4f3558ea582a85b08e026bec6e0a4" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve```

Screenshot from 2024-10-25 11-02-10

**3. A workflow_dispatch on a branch with name release-* .** This will look identical to what is shown in case 1. 
Some considerations for Case 3:

- When we trigger this workflow_dispatch we should specify we want the image to be tagged "prod". I've updated the help-text displayed when you trigger a workflow_dispatch to help with that "Docker tag for the image, default is dev. Use prod for a release branch fix'. 

- This will require us to name the release branches like this: release-vX.Y.Z  that seems to be the standard way to do it though. 

**4. Any other workflow_dispatch -** aka working on a feature branch and want to test
Can test locally with: 

```SWG_HUGO_ENV_ARG="some-feature-branch,fbabebb3dfb4f3558ea582a85b08e026bec6e0a4" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve```

Screenshot from 2024-10-25 11-03-55

**5. If no param provided (aka not a GH action but local development)** it will show up as follows:

```./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve```

Screenshot from 2024-10-25 11-54-54

### CODE REVIEW

File: scripts/dockerbuild [hunk 1]
```diff
@@ -28,6 +31,10 @@ fi
 if [[ "$1" = hugo ]]; then
     _DEFAULT_IMAGE=ghcr.io/scilifelabdatacentre/swg-hugo-site
     _DEFAULT_DOCKERFILE=docker/hugo.dockerfile
+    declare -a _BUILD_ARGS
```
kwentine, line 34: The `_BUILD_ARGS` declaration could be moved at the top scope. But more importantly, perhaps: should we provide a default value ? For example:
```bash
"HUGO_ENV_ARG=${SWG_HUGO_ENV_ARG:-$(git rev-parse HEAD)}"
```
I also like `git describe` but it shoud be used with some care here since it fails without a least one tag earlier in the history.

scripts/dockerbuild
```diff
@@ -5,6 +5,9 @@
 # By default, the host user ID and group ID are forwarded through
 # build arguments. To disable this behavior and use the dockerfile
 # defaults, set the SWG_DEFAULT_USER environment variable to 1.
+#
+# The hugo image can be built with the SWG_HUGO_ENV_ARG environment variable
```
kwentine, line 9: I would suggest a more targeted name like `SWG_VERSION_INFO` (or something similar) And the corresponding modification of `HUGO_ENV_ARG`.

docker/hugo.dockerfile
```diff
@@ -14,17 +18,26 @@ RUN wget --quiet "https://github.com/gohugoio/hugo/releases/download/v${HUGO_VER
...
 
+# GH actions triggered runs will overwrite HUGO_ENV_ARG 
+ARG HUGO_ENV_ARG="Local dev build"
+
+# Set the combined environment variable for hugo build
+ARG ENV_INFO="${HUGO_ENV_ARG}|${JBROWSE_VERSION}"
+
 RUN mkdir /target && \
-    hugo -d /target
+    hugo -d /target --minify -e "${ENV_INFO}"
```
kwentine, line 28: I am a bit worried about using the `-e` option, since it may be relied upon by `hugo`. For the selection of the configuration file for example, or the [hugo.IsDevelopment](https://gohugo.io/functions/hugo/isdevelopment/) function. Do you think it would be possible to pass regular environment variables to the build and then use [os.Getenv](https://gohugo.io/functions/os/getenv/) This would allow you to pass as many as you please, an ease the parsing logic in `version_info.html`

kwentine (commented):
This looks great, and the PR description was 👍🏻 ! Can't believe we finally stepped into the matrix. I have not finished my review yet, but I wanted to get started with a few comments before we meet tomorrow :)

docker/hugo.dockerfile
RMCrean: Great idea, I copied the approach by the pathogens portal with using -e, but I really like this idea. As you say, it will simplify the logic in `version_info.html `a lot. 
I will work on making this change and let you know when it is done :)

scripts/dockerbuild
RMCrean: I've moved the `_BUILD_ARGS` to the top. 

The default currently says "local dev build" to reflect that you might have some uncommitted changes you are testing.

scripts/dockerbuild
RMCrean: I've now updated the names/used different names as we have different variables. I'm not attached to them though so feel free to commit alternatives if you prefer.

RMCrean:
hey @kwentine, thanks for your help. I think this is now ready for a 2nd round of review. I've implemented your idea about passing regular environment variables and that has definitely made the logic in `version_info.html `a lot smoother. 

If you want to, please free to add a commit to make some changes. 

P.S. I checked the "-e" flag issue of whether we need to specify it and it is [as we expected that as long as you just use "hugo" and not "hugo server" then you get "production mode"](https://gohugo.io/functions/hugo/environment/)

#### New commands. 
##### local dev build
`./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve`

##### release (from creating new release or workflow_dispatch on release branch)
`SWG_HUGO_PORTAL_VERSION="V1.0.1" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve`

##### build from push to main
`SWG_HUGO_GIT_BRANCH="main" SWG_HUGO_GIT_SHA="1b8a04baddb27679f4d7ef60f64f7d110f6e9a4e" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve`

##### build from workflow dispatch
`SWG_HUGO_GIT_BRANCH="some-feat-branch" SWG_HUGO_GIT_SHA="1b8a04baddb27679f4d7ef60f64f7d110f6e9a4e" ./scripts/dockerbuild hugo ; SWG_TAG=local ./scripts/dockerserve`

kwentine:
> I think this is now ready for a 2nd round of review.

Thank you for the nice recap and walk through of the different scenario! Starting a new review.

kwentine:
Some comments:
- 298a3d2 Found out that build arguments get passed to the `RUN` environment, making `export` redundant
- ec7ef69 I saw no real need to pass contextual arguments to `dockerbuild` using environment variables
- e93723a A bash trick: redirections can be applied to compound commands as a whole
- a993aa4 Parameter expansion is quite powerful, and here more efficient (avoids launching a sub-process)
- 4197a42 to 6e55df6 Give Hugo the unprocessed CI context and deal with it in the template

kwentine:
@RMCrean I have extracted the version-parsing logic to partials, to make it convenient to re-use in #69
If you are fine with the state of things, I'll let you merge this PR 🚀 :) Otherwise push-ahead, I am done!

RMCrean:
Great, thanks for that. I'll merge now and then update the Hugo image shown on the dev website :)

PR #63, opened 2024-10-29
Test of rendering secondary assemblies in separate views

brinkdp:
The case of adding support for displaying separate secondary (organelle) assemblies in their own view(s) within a JBrowse session has been on my back-burner for a long time and it's about time to test it out. At the moment, we have encountered one real case: *Parnassius mnemosyne* (EB-97), but I would not be surprised to see more of them coming in the future. 

This PR contains a working example for how we could implement support for multi-assembly sessions in the Genome Portal. The experiment did however raise two questions:

1.
Due to the structure of the two main nucleotide repositories, the mitochondrial assembly is not available via ftp from either ENA or NCBI, only via their APIs. From what I understand, our current `curl` setup is not able to download files from URLs that call on the ENA API: for example if a url in `config.yml` points to `https://www.ebi.ac.uk/ena/browser/api/fasta/GCA_964186615`. After reading the suggestions in the [ENA API documentation](https://www.ebi.ac.uk/ena/browser/api/swagger-ui/index.html#/ ), I am able to download the file using this command:

```
curl -X 'GET' \
  'https://www.ebi.ac.uk/ena/browser/api/fasta/GCA_964186615?download=true&lineLimit=0&gzip=true' \
  -H 'accept: text/plain' \
  --output GCA_964186615.fasta.gz
```

Since organelle assemblies tend to be small (this example is ~4.5 kb when compressed), a temp solution is to host the file ourselves for now. But eventually, it would be useful if `make`could support these cases. What do you think @kwentine ?

2.
Running `./scripts/dockermake --test SPECIES=mito_clouded_apollo` on the 2da2af1 examples result in errors from `jbrowse add-track`. The cause of this is completely reasonable: once there are multiple assemblies, `add-track` requires `--assemblyNames` to specify to which of the assemblies to assign the track.

I fixed this in 563b4af by some reverse-engineering. I manually added the mito assembly and annotation track to `tests/data/mito_clouded_apollo_workaround/config.json`, and, once it functioned satisfactorily, copied over the full content to `tests/config/mito_clouded_apollo_workaround/config.json`. I then set every single track in `tests/config/mito_clouded_apollo_workaround/config.yaml` to `addTrack: false` to  circumvent `--assemblyNames` errors, while still triggering the downloads.  In this way, the initial config.json passes (more-or-less) unaltered through the JBrowse CLI calls of the makefile. 

I am happy to transplant this work-around to the main parnassius_mnemosyne configs, similar to how we decided to keep a high degree of hand-crafting for the GWAS plugin-dependent tracks. Do you think is the way to go, or is better to implement a key in config.yaml that passes `--assemblyNames` options to `jbrowse add-track`?


kwentine:
> From what I understand, our current `curl` setup is not able to download files from URLs that call on the ENA API

In 6e6bbd7 I added a minimal configuration with the desired ENA API URL and `./scripts/dockermake --test SPECIES=test_ena_api` seems to handle it well :)

kwentine:
> I fixed this in [563b4af](https://github.com/ScilifelabDataCentre/genome-portal/commit/563b4af8bf89c1694fde490d15451d90d865f783) by some reverse-engineering.

I have a hard time understanding the workaround. I think you'll have to show me on Zoom or something (or maybe it's just my Friday brain that is a bit dull)

brinkdp:
Thanks, it really seems that the part that I was missing was the declaration of the `fileName` :sweat: 

I admit also had some trouble today getting the work-around to run at first. I guess there were things that I forgot to write down and/or implement last time. I'm sorry if this was an inconvenience for you! For instance, the call for `http://fixtures/mito_clouded_apollo/OZ075093.1.fna.alias` gave an error in JBrowse when used with `./scripts/browse`. 

Anyway. The new logic works nicely, big thanks for that 🙏 

I added some flair to the example while I was testing it out. With 10c8607 and e8d5069, the alias file will now be downloaded and called from `tests/data` like the other files (circumventing the JBrowse error for this test case), and defaultSession is styled with the envisioned look. 

For me, the following gave reproducible results while I was working on testing and styling the example:

```
if [ -d "tests/data/mito_multi" ]; then rm -r "tests/data/mito_multi"; fi
SWG_TAG=local ./scripts/dockermake --test SPECIES=mito_multi 
docker stop jb2-mito_multi; SWG_TAG=local ./scripts/browse tests/data/mito_multi
```

Now after doing some hands-on work with this and reading the documentation you linked, I have come to like the YAML documents implementation. When considering future code reuse cases, however, I think that the nested YAML list approach a bit more intuitive. But I don't know if the preference of one person - myself - is enough to support a big rework. I would be happy with the "documents" design, if you also think that's a good way to go.

kwentine:
And a question: do you wish to add the mitochondrial assembly in this PR, or would you prefer me to merge these technical underpinnings first, and add the assembly tracks and configuration in a separate PR ?

brinkdp:
> the strategy is brittle: it worked because the alias happens to have .fna in its filename.

Ah, fair point. It happens to be the case that the alias files are generated from the assembly file downloaded and processed by the make file, meaning that all alias names generated for the portal will include the `.fna.alias` extension. But in a way, I see that my approach was partly a lucky shot, haha. Using e.g. a Google Drive is better, I completely agree.

However, the drive calls does not work for me when I inspect the resulting JBrowse sessions. The tracks are empty (Fig. 1), which is the tell-tale sign that something is wrong with the aliases. Does it work for you?

Screenshot 2024-11-28 at 10 02 02
_Fig. 1. Empty tracks when using the Google Drive calls implemented in 0d6cb26_

Screenshot 2024-11-28 at 10 06 46
_Fig. 2. Desired results, previously achieved with the hack that downloaded the aliases_

I've tried to troubleshoot this but to no avail. I changed the alias call back to the raw GitHub file for `CAVLGL01.fna.alias` to ensure that the way I was testing was not causing the issue with the remote file (`dockermake --test` and `browse` as described in my last post in the thread). I checked that the files were properly formatted with tabs, which has been a source of error for refNameAliases in the past (e.g. d3b6670). The trick with downloading the file from GD as a track and calling it from the data directory (Fig. 2) also work after changing the url to the format of https://drive.google.com/uc?export=download&id=FILE_ID. 

So that left me thinking that there is something that does not work with calling the remote GD file url directly with the alias key. Putting the `uc?export=download&id=FILE_ID` version of the url in the `aliases` field gives a JBrowse error, which did not surprise me. But it tells me that the fact that `https://drive.google.com/file/d/FILE_ID/view?usp=sharing` gives no JBrowse error is indicative of the being file is fetched as intended but that there is something wrong with its content, resulting in the empty tracks. This is where I got stuck in my troubleshooting. I wonder if the way we access the GD url results results in a modified file and not the raw file that we need for JBrowse?

> And a question: do you wish to add the mitochondrial assembly in this PR, or would you prefer me to merge these technical underpinnings first, and add the assembly tracks and configuration in a separate PR ?

I think we can merge this once we sort out how to use the drive, and then I can make a separate PR for updating Clouded Apollo.

kwentine:
> Using e.g. a Google Drive is better, I completely agree.

I have a nagging feeling that this is perhaps not the best solution, but can't really pinpoint whence it comes (I mean aside from the little detail that it doesn't work 😅 ) If Google Drive and Nexcloud are inadequte, and Figshare a bit unwieldy, do you see other options we could try ?

From studying your valuable flowchart I realize that we are routinely generating ad-hoc aliases. Would you like that to be incorporated as another `Makefile` recipe ? Roughly, a `make aliases` that would call `get_aliases_from_ENA_fasta.py` as part of the build step, for all the ENA assemblies ? Send me a calendar invite if you'd like to discuss it.  
 
> Does it work for you?

I suspected something was wrong when I saw blank annotation tracks display...


> I've tried to troubleshoot this but to no avail [...] The trick with downloading the file from GD as a track and calling it from the data directory (Fig. 2) also work after changing the url to the format of https://drive.google.com/uc?export=download&id=FILE_ID.

Thanks for your leads and helpful troubleshooting hypotheses. I will investigate further.

> I think we can merge this once we sort out how to use the drive, and then I can make a separate PR for updating Clouded Apollo.

Sounds like a plan 👍🏻

kwentine:
After unsuccessful attempts to host files on Nextcloud and Google Drive, I reverted your astute solution enabling aliases in the `fixtures` folder (7889fb1) I hope we can work together to include alias file generation as part of or build recipes.

brinkdp:
Thanks for doing the extra round of testing! Too bad it did not work. At least we know that these two cloud storage services were not suitable for our needs.

> Would you like that to be incorporated as another Makefile recipe ? Roughly, a make aliases that would call get_aliases_from_ENA_fasta.py as part of the build step, for all the ENA assemblies ?

I think this is a great idea, let's work together on this! I assume including it in the build process opens up for storing the generated aliases on the persistent volume, thus providing a solution to this current alias issue?

Some context: my initial idea early in the project was that alias files would only be needed for exceptional cases where the data would be fetched from multiple repositories. In hindsight, the opposite turned out to be the case: only 1 of 6 launch species do not need an alias file. Thus I routinely began to generate aliases for all the ENA assemblies and this is what is reflected in the flow-chart. Having it as part of a build recipe would be an asset, indeed.

kwentine:
> I assume including it in the build process opens up for storing the generated aliases on the persistent volume

Exactly!

> let's work together on this
 
I have created an [issue](https://scilifelab.atlassian.net/browse/EB-396), and will start by examining your script and reach out directly if I have questions. When I have strategy to plug it into the  `make` machinery, I'll open a PR with a draft implementation and put you as a reviewer so that we can refine it together.

In the meantime, is it OK for you if I merge this PR ?

brinkdp:
Excellent, sounds like a plan! Yes, please go ahead and merge this PR.

> 🚀 ✖️ 🫏 (this is a quizz)

The biologist in me want to interpret this as a the hybrid name for a cross between a rocket and a donkey. So an android mule? 😜 But I guess it's short-hand for a stubborn issue blocking deployment?

PR #66, opened 2024-11-05
EB-280. Update krill according to suggestions from PI

brinkdp:
This PR adds new updates to the Krill pages based on feedback from the submitting researcher.
- New photo
- Some rephrasing of the species description (I wrote the original draft based on the dataset-associated publication)
- Contact info to the PI
- Listing of new data tracks for the sake of completeness. NOTE! No new tracks were added to the JBrowse 2 config (this was not requested)

In other words, this is mostly a straight-forward content update. The last bullet can be discussed, however. The template text above the data table states that the table only lists data that is displayed in the Genome Portal, but here I deliberately deviated from that. My reasoning is that we have been promoting that we will offering high customization of the pages based on the needs of each research group, and we were asked by the PI to put in these tracks in the table. And it aggregates more data tracks, which is also in line with the scope of the portal.

I put `(currently not displayed in the Genome Portal)` for the tracks that are not configured for JBrowse 2. I think that will be enough for the launch. I don't think we need to change the body text above the table at the moment, but I'm happy hear your thoughts!

### DIFF
--- a/hugo/assets/meganyctiphanes_norvegica/data_tracks.json
+++ b/hugo/assets/meganyctiphanes_norvegica/data_tracks.json
@@ -15,7 +15,7 @@
     "firstDateOnPortal": "15 October 2024"
   },
   {
-    "dataTrackName": "Protein coding genes (research group's version, used on the Genome Portal)",
+    "dataTrackName": "Protein-coding genes (research group's version, used on the Genome Portal)",
     "description": "Functional annotation of protein-coding genes with non-redundant isoforms. Missing stop-codons have been added if detected in the reference genome assembly",
     "links": [
       {"Download": "https://figshare.scilifelab.se/ndownloader/files/40569599"},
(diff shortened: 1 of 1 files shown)


hugo/content/species/meganyctiphanes_norvegica/_index.md
```diff
@@ -31,9 +29,9 @@ initialZoom: 1
...
+The wide distribution and thermal tolerance of *M. norvegica* make it a powerful model to study ecological adaptation to different marine climates and rapid ocean warming. A challenge with genomic analysis of krill species is their huge genome size (19 Gb for *M. norvegica*) (Choquet et al. 2023; Jeffery, 2012; Unneberg et al., 2024). While most genetic variation appears to be widespread among individuals throughout range of the species, researchers have also identified many signatures of local genetic adaptations in genes involved in responses to light and temperature (Unneberg et al., 2024). In particular, krill from the Atlantic Ocean and the Mediterranean Sea are distinctly different at hundreds of genes. Such genetic variation could be crucial for allowing the krill to adapt to continued climate change and promote reliance of krill populations in a warmer ocean.
```
apfuentes, line 34: In [individuals throughout range of the species] add "the" before "range".

apfuentes (approved):
Excellent job @brinkdp, I only spotted a minor typo (mentioned as a comment).

PR #70, opened 2024-11-22
EB-315: Create GH action for playwright tests

RMCrean:
Hello, 

This PR is to add a GH action that runs on a PR to main only if relevant files have been edited.  

One thing I'm uncertain about is how best to handle a flaky test (`test_table_links`  located at: `playwright/tests/test_species_download_page.py` ). This test checks each link in the Links column of the table on a species download page to see if they look "okay". For a download link (where the problem is), the test waits to see if the download begins within 30 seconds of clicking the link (default timeout from playwright). 

Locally this works fine - except for one file located in ( http://localhost:8080/linum_trigynum/download/) which @brinkdp  already knows about and is already working on trying to get fixed. 

With the GH runner it fails for multiple download pages (4): See the failure report here: https://github.com/ScilifelabDataCentre/genome-portal/actions/runs/11854290028/job/33036142653  My understanding is that this is because the test is flaky with the GH runners. 

My current solution is to just skip the test when run with GH actions using the `-k` flag in pytest, as I think it is better to not create misleading fails and it is also an easy solution. This is perhaps also nice considering the test will fail currently even if the test is not flaky because of the link that we know is broken.  

Are you both okay with that? Happy to try out some suggestions ideas on what we could do.


kwentine (approved):

RMCrean:
So in attempt to debug the cause of the failure (commit 44d28827fe500ff95b413aac1d92ee77acc1ad6c), the failure seemed to go away... 

I've now reverted that commit and then added a simple check to not fail on the link we already know is bad. That seems to work nicely. 

Perhaps Figshare was having an issue the moment the initial tests were ran or something. I think we can merge as is and if we see a failure relating to the `test_table_links` function we can be a little bit sceptical about if it is a real error or no (easy way to validate will be to run the tests locally). 

Does that sound good?

brinkdp (approved):
Thanks for this work, looks neat! 

> Perhaps Figshare was having an issue the moment the initial tests were ran or something. I think we can merge as is and if we see a failure relating to the test_table_links function we can be a little bit sceptical about if it is a real error or no (easy way to validate will be to run the tests locally).

I think your reasoning sounds good and pragmatic. I am curious to see if this particular Figshare-related error will reoccur. `test_species_download_page.py`did not throw any errors for me when I ran it locally today, at least 🤞

RMCrean:
> I think your reasoning sounds good and pragmatic. I am curious to see if this particular Figshare-related error will reoccur. test_species_download_page.pydid not throw any errors for me when I ran it locally today, at least 🤞

Good to hear! 

Thanks both, I will merge now then :)

PR #71, opened 2024-11-26
EB-293: Decide when links should and should not open in new tabs 

RMCrean:
Hi! 

I've made some edits to the Hugo content files to change all internal links (excluding opening the genome browser) to open in the same tab by default. I have not edited any external links that were already set to open in a new tab by default. 

Looking at various UX guides/articles online it seems to be preferred to have as few links as possible open in a new tab by default and there is some debate about whether external links should by default open in a new tab too. Regardless, I think this first commit is a good starting point (and potentially "good enough" already). 

Feel free to state your opinions or make a commit to this branch. We can also discuss this briefly in the meeting later today


apfuentes (approved):
Hi @RMCrean, it sounds good to me. Personally, I prefer to open on a new tab a link that directs me to another page, but I am totally fine if this is not the preferred case :). Please go ahead with the PR.

RMCrean:
Great, thanks for both taking a look! I will merge now.

> I realized that I had not considered how opening of links affect the genome browser page. I clicked around a bit, and it seems that the state of the JBrowse session is cached, so if a user accidentally clicks on an internal link while working in the genome browser, they can return to the state of the session by using the back-button the their browser. That's all good, I think. But if they go back to the species page and click on the genome browser button, the defaultSession will be launched, loosing the previous state. This is the expected behavior, but I wonder if it is clear to the users. Anyway, I have no good answer for this myself, so I'm throwing it out there for discussion.

Thanks for bringing this up, I can confirm I experience the same thing and I'm not sure how easy this would be to "fix" or if we should either if it is the expected behavior?

PR #69, opened 2024-11-14
EB-282: Add copy-to-clipboard button to facilitate copy of citations

RMCrean:
Hey, 

This PR is is for adding copy-to-clipboard buttons for the citations and any code blocks. This PR doesn't include embedding the version info directly into the citations but it should be fine to include that at a later stage (in the shortcode) once ready (dependent on PR #62 ). 

I followed this [guide which summarizes quite nicely the steps required](https://logfetch.com/hugo-add-copy-to-clipboard-button/). There are some differences but the general idea is still there. 

**Adding a copy-to-clipboard button :** 
1. For a citation, Wrap the citation in the hugo shortcode `citation_block.html `, see `hugo/content/citation/_index.md` for examples. 

Screencast from 2024-11-14 15_05_46

2. For code blocks, use standard markdown format, i.e.:  ``` 
Code blocks already marked with this are automatically recognized and provided with a copy-to-clipboard button. 
See examples of what this looks like on: `http://localhost:8080/contribute/supported_file_formats/`

**Notes:** 
1. Initially I tried using a shortcode to create the code blocks (as done for the citations). This caused the markdown linter to go crazy and seen as how it didn't provide any extra functionality/benefit I think it is best to stick to using: ``` for code blocks. 
2. Using the shortcode `citation_block.html ` is a bit of a hack. It essentially puts the citation text inside the same HTML structure that a code block has. This enables it to be recognized by `hugo/layouts/partials/copy_to_clipboard.html` 
3. I put the copy-to-clipboard button outside the code/citation block. I read somewhere that this is good to avoid putting it inside the block as it can end up sitting on top of the code/text if it makes it all the way to end of the line. 
4. The styling of the code block and citations blocks are identical with the exception that wrapping lines is allowed in the citation blocks.
5. Updated Hugo version to take advantage of https://gohugo.io/functions/strings/trimspace/

PR #72, opened 2024-11-27
EB-371: Add copy to clipboard for species pages

RMCrean:
Hi! This PR is about extending #69 with copy-to-clipboard blocks for the species pages:

**Other changes:**
- replaced the html used in the citations of the species `_index` and `assembly` markdown files to markdown to hopefully make it a little easier to read.
- Updated the `add_new_species.py` templates files to this new setup. 

**Build vs Current date:** 
Part way along the way I realized I had set that the date accessed variable in the citation blocks (both here and on the cite us page) using the Hugo build date and not the actual/current date. To deal with this I've created a little Hugo shortcode `hugo/layouts/shortcodes/todays_date.html` and a js script `hugo/assets/js/todays_date.js`. The logic is essentially to create an empty span element with class "todays-date" that is replaced by the current date with JS. This doesn't impact what the copy to clipboard button actually copies. 

I also use this approach for the in text mention of the date accessed which is the first blue "circle" in the pic below:

Screenshot from 2024-11-27 10-25-57

It seems to work correctly. I will also now trigger an image build with workflow_dispatch and update the dev website with it. Just to sanity check tomorrow that all the dates correctly display...

### DIFF
--- a/hugo/assets/js/todays_date.js
+++ b/hugo/assets/js/todays_date.js
@@ -1,0 +1,12 @@
+/*
+Get the current date and set it in any span element with class "todays-date"
+Format looks like: November 27, 2024
+*/
+const options = { year: 'numeric', month: 'long', day: 'numeric' };
+const today = new Date();
+const formattedDate = today.toLocaleDateString("en-US", options);
+
+const dateElements = document.getElementsByClassName("todays-date");
+for (let element of dateElements) {
+    element.innerHTML = formattedDate;
+}
--- a/hugo/content/species/clupea_harengus/_index.md
+++ b/hugo/content/species/clupea_harengus/_index.md
@@ -47,7 +47,7 @@ Write some content here in markdown for the introduction tab of the species page
 
 [EDIT]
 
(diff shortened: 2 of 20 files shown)


apfuentes:
@RMCrean thanks for implementing this feature, from the shared images, it looks good! Not sure I can look at the code any time soon unfortunately.

brinkdp (approved):
Neat!
I played with sending different values to HUGO_GIT_REF_NAME and the result was correct in the pages and in the clip-board content. Looks good from a user point-of-view. Nicely done!

As for the js, I did not look to closely at that, but it seem to get the work done. Do I understand it correctly that the built-in feature for current date in hugo that we were talking about in another PR (I forget which, perhaps the archetypes one) was only considering the build date?

brinkdp:
>Probably the archetypes one yes as we were talking about [lastmod](https://gohugo.io/methods/page/lastmod/) during that one. With Hugo you could get both current (i.e. build time) and past dates (e.g. something like the last time the file was modified). But as Hugo's code is run at build time it cannot perform future calculations, so you can't write Hugo code that will be run when the user visits the website 1 week from when the site was built, so that's why we need to use JS in this case (sorry in advance if you already knew that part and were just asking about what the Hugo method was called).

Thanks for the clarification! This was indeed what I suspected, but great to get the extra explanation

PR #73, opened 2024-11-27
EB-390: Github contrbution guide

kwentine:
This PR drafts a Github contribution guide for users who would be willing to submit new species through pull requests.

I could be paired with an adequate pull request template. 

I plan to add subsequent sections for undaunted users who would like to get started with their species content
using `hugo new` or `add_new_species.py`.

Finally, I used the [AsciiDoc](https://docs.asciidoctor.org/asciidoc/latest/lists/ordered/) format that was praised at PyCon to give it a try. If you prefer sticking to Markdown, it will give me a nice opportunity to try Pandoc 🙂 

All remarks welcome of course, this is an experiment!

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -1,0 +1,79 @@
+= Contributing
+
+This guide describes how to submit a new species for the
+https://genomes.scilifelab.se[Swedish Reference Genome Portal]
+
+In the following examples, we will assume that you (Github user
+`insidethebox`) want to add the
+https://en.wikipedia.org/wiki/Yellow_boxfish[Yellow boxfish]
+(described as _Ostracion cubicum_ by Linné in 1758) species to the
+Genome Portal.
+
+== Prerequisites
+
+Start by
+https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#forking-a-repository[forking]
+the
+https://github.com/ScilifelabDataCentre/genome-portal/fork[genome-portal]
+repository from your Github account.
+
+Clone the repository to your computer, and create a branch to work on. Name the
+branch after the scientific name of the species:
+
+```
+git clone git@github.com:insidethebox/genome-portal.git
+cd genome-portal
+git checkout -b ostracion-cubicum
+```
+
+== Reference genome and gene annotation
+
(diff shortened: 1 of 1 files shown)


RMCrean (approved):
LGTM :)

One thing that could be added is to encourage them to allow us to commit to their forked branch which they can do by ticking a button in the PR interface: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

apfuentes:
@kwentine great initiative! I wonder if it would be good to have a "dev" branch to which user's PRs are first merged and tested, and then if everything looks good (pass our review), then the PR is merged to main?

kwentine:
>  I wonder if it would be good to have a "dev" branch to which user's PRs are first merged and tested

Thanks for bringing up that point. 

Since we can deploy any branch for testing purposes on our development portal `https://swedgene.scilifelab-2-dev.sys.kth.se/`, I think both review and preview can occur directly on the PR branch. In other words, nothing additional could be tested by introducing a `dev` branch. But curious to hear what @RMCrean and @brinkdp think.

kwentine:
> One thing that could be added is to encourage them to allow us to commit to their forked branch

Great suggestion, thanks. Mentioned it in cd07bb4

RMCrean:
I agree that I don't think we need a `dev` branch for this case.

brinkdp (approved):
Apologies for being late on this review! Looks great, thanks for the initiative.

Good work on this documentation. I think it is clear and reads well, assuming the target audience is a person who has some familiarly with GitHub. The .adoc format seem interesting. I did not look to closely into it, but I am curious to hear what benefits you think it brings compared to markdown? 

I agree fully with the idea of having the fork PR being open for collaboration, thanks for adding that. In my opinion we will probably always need to do some final touch-ups or troubleshooting in each case, even if we in a near future hope to have helper scripts for populating defaultSession and the three species tabs that the submitter can use for their PR 😄 

I also agree that we do not need a dedicated `dev` branch for this. The current local and dev portal options are fully suitable for our testing needs. 

I recall that you mentioned at some point that it would be possible to make a PR template for contributions? Is that a line we would like to pursue?

PR #75, opened 2024-12-03
EB-314: Make pagination possible for species search

RMCrean:
Hey! 

I seem to have pagination setup now with pure vanilla JS (nope, not http://vanilla-js.com/ ;) ), it was a great learning experience :).  
[The dev site also currently displays what it looks like](https://swedgene.scilifelab-2-dev.sys.kth.se/): 

Screenshot from 2024-12-03 11-43-01

### Overview:
- Whether to paginate is controlled by the number of species and the `maxSpeciesPerPage` param in the `hugo/config.yaml`. If pagination is not applied (e.g. `maxSpeciesPerPage` > 6, then the search etc.. still works and so do the playwright tests). 
- A JSON file containing info needed to build each species card is created at build using the `hugo/layouts/index.json.json` template file. This is loaded upon loading the page. The `hugo/config.yaml` has some params for this file. 
- The `hugo/layouts/index.html` decides whether to add the HTML for the pagination buttons based on the `maxSpeciesPerPage` param. 
- The script `hugo/assets/js/species_cards.js` handles the filtering/sorting/display etc.. 

### With `hugo/assets/js/species_cards.js`:

**On load:**
1. JSON data with all species attributes fetched ('index.json.json' defines the structure of the data)
2. Species cards populated with this data

**Once loaded:**
- Event listeners created for search box, dropdown and pagination (if pagination exists).

### Playwright: 
- Tests work with and without pagination, when no pagination present, some tests return early.  

### Before merging:
- Currently `maxSpeciesPerPage` is set to 2, so you can see what it looks like with pagination. I think we should set this to at least 8 before merging, otherwise it looks a little weird.


hugo/layouts/index.html
```diff
@@ -52,168 +57,58 @@ <h2>Explore species</h2>
...
 
+<!-- Add pagination if numb of species cards is > than number of species per page
+     numb species per page set in main config.yaml file -->
+{{if gt $numbSpecies $numbSpeciesCards }}
+    {{ partial "pagination.html" (dict "numbSpecies" $numbSpecies "numbSpeciesCards" $numbSpeciesCards) }}
+{{end}}
+
+
+{{ $speciesCardsJS := resources.Get "js/species_cards.js" | fingerprint | minify }}
+<script type="module" src="{{ $speciesCardsJS.RelPermalink }}"></script>
```
RMCrean, line 111:
```suggestion
<script src="{{ $speciesCardsJS.RelPermalink }}"></script>
```
This is a standalone script so I shouldn't have this here

apfuentes:
@RMCrean the pagination looks great! The screenshot shared shows two cards. Would it be possible to set that 4 cards are shown by default, when card filtering is not triggered by the search bar?

RMCrean:
> @RMCrean the pagination looks great! I suggest to set that 4 cards (or multiples of 4) are shown by default, when card filtering is not triggered by the search bar, so the webpage looks balanced.

Thanks @apfuentes! :) Yes the current setting of 2 per page is temporary (just to help illustrate the pagination works). I definitely agree with the multiple of 4 idea and personally think before 9 species we should still not have any pagination (open for alternatives). I suggested in my PR comment above that once we're happy to merge we set the `maxSpeciesPerPage` inside `hugo/config.yaml` to 8, which would effectively do that. Does that sound like a good point to paginate to you?

kwentine (commented):
Fantastic work, and all the better if it was instructive!

I'll throw in a couple of ideas that come to mind for this first round of review. And submit them as pointers that together form a sort of riddle:

- The HTML [<template>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) element
-  Embedding data or text in a [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#embedding_data_in_html) tag

And a function to consider:

```javascript
function render(template, context) { 
  return template.replaceAll(/\{(.*?)}/g, function(match, captured) {return context[captured]}) 
}

render("I am a {animal} and relish {food}", {animal: "squirrel", food: "hazelnuts"})
```
Do you see where I am going ? If not, let's talk about it on friday :)

kwentine:
I unfortunately won't have the time to do a detailed code review this morning, so I will postpone it to after our discussion on Friday at in the meantime share another idea (elaborating on the above).

One way to think about the problem is to model the interface state as a data structure:
```javascript
state = {
  species,
  currentPage,
  pageSize,
  sortKey,
  sortOrder,
  searchString
}
```
Given this information, a root node and a species template, we have all we need to render the view (Iet's set aside the nav controls for now). In other words:
```javascript
function render(state, root, template) {
  // Clear the content of root
  // Render species templates under
  // Append to root
}
```
Then a perhaps slightly inefficient but very simple way to reason "functionally" about the problem: UI events update the state, and every time the state changes we call `render()` Then a lot of the logic is represented by `state -> state` functions that are easy to test and reason about. 

A more sophisticated introduction to these ideas: the [Redux three principles](https://redux.js.org/understanding/thinking-in-redux/three-principles)

And a final suggestion: perhaps you have already implemented it, but you could consider implementing "debouncing" of the search : do not trigger the search on every keystroke, only when the user has finished typing...

RMCrean:
Thanks for the ideas here and the chat on slack, I see what you mean and am up for having a go implementing a version of this. I'll mark this PR as a draft for now (and make some changes), and then when I think I've made some progress I can request another review.

I think I'll focus first on the HTML <template> inclusion (which should potentially remove the need for the JSON), and then creating a global state object and then building out the render function and we can go from there. 

> And a final suggestion: perhaps you have already implemented it, but you could consider implementing "debouncing" of the search : do not trigger the search on every keystroke, only when the user has finished typing...

I hadn't considered this no, I will look into it.

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+const SORT_OPTIONS = {
+    LAST_UPDATED: 'lastUpdated',
+    A_TO_Z: 'alphabet',
+    Z_TO_A: 'revAlphabet'
+};
+
+const paginationItems = document.querySelectorAll('.pagination .page-item');
+const noResultsCard = document.getElementById('no-filtered-card');
+const PAGINATION_EXISTS = paginationItems.length > 0; // if not enough species, there will be no pagination yet.
+const CARDS_PER_PAGE = parseInt(document.getElementById('card-container').dataset.numbCardsPerPage);
```
RMCrean, line 27: A perhaps minor but general question on when to capitalize constants (not just `const`s but things that are not supposed to ever change) in JS. 

I feel like using all caps for: `SORT_OPTIONS` and `CARDS_PER_PAGE` makes sense but not sure if it would make sense to name `paginationItems` or `noResultsCard` which also never change (and should never change) with all caps. Do you have an opinion on that?

RMCrean:
Hey Quentin, I think I'm ready for round 2 now. I've now adapted the code quite a bit. 

I really like the `<template>` element, so I decided to use it for both the species cards and for the dropdown button active text (to help further reduce DOM manipulation with JS) :) . 

**Now on page load:** 
- The info needed to search and order the species cards is taken from each <template> card.
- The render function is run to display cards. 

**Event listeners:** 
- Function in a similar way as before with the exception that I have now added debouncing to the species search. Googling says a 200 ms delay is lagom and that seemed okay to me. 

**Other Notes:** 
- In your example code you passed arguments into your render function whereas I have just written the functions to directly interact with the `state.[object_name]`. I suppose there are pros and cons to both (the code I have written is more tightly coupled because of it I suppose). I'm happy to change it to the other way round if you prefer. 
- I decided against using a `<script>` tag in the HTML so that we can slowly but surely work towards the content security policy JIRA task, but the need for the JSON species file is gone now at least :).   

P.S. I have a minor question for you about variable naming (it doesn't feel as obvious to me when to use ALL_CAPS and not as compared with Python) - it's a code comment on this PR.

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+Handles searching/filtering, ordering and pagination of the species cards on the home page.
+This script works with 'index.html' inside the hugo/layouts folder.
+
+On load:
+1. Species attributes needed for searching fetched from the <template> cards built for each species.
+2. Determine if pagination exists (decided by Hugo based on number of species and 'maxSpeciesPerPage' in 'hugo/config.yaml').
+3. Run the render function to display the species cards that should be shown.
+
+Once loaded:
+- Event listners created for search box, dropdown and pagination (if pagination exists).
```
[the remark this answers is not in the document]
kwentine: This is a good occasion to sum up the spirit of my remarks below. My ethereal ideal would be something like:
```javascript
function searchHandler(query) {
  // Only event handlers mess with the global state variable
  state.searchString = query
  //Rendering functions get a copy of the state and the root node as arguments
  render({...state}, root)
}

function render(state, root) {
  // Sorting and pagination function work on data structures passed as arguments
  // This makes them easier to unit test
  const species = filterAndPaginateSpecies(state)
  root.innerHtml = ''
  renderSpeciesList(species, root)
}
```
Thus, aside from the event listeners that access the global state, all our functions take data structures (and a root node, for rendering functions), which makes them convenient to test.

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+    });
+    state.numbMatches = templates.length
+
+    renderSpeciesCards();
+}
+
+
+// Sorting functions below
+
+function stringToDate(str) {
```
kwentine, line 63: Here is another weak point where changing the date format in Hugo would break the search. The very hypothetical day we decide to localize the site for US viewers for example. To be contrasted with sorting based on a `date` JSON field which would be unambiguously serialized.

But these are again lofty design considerations, I would be fine to stick with this if we forsake the JSON approach.

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+            filteredSpecies.sort(sortLastUpdated);
+        } else if (state.sortOrder === SORT_OPTIONS.A_TO_Z) {
+            filteredSpecies.sort(sortAlphabetically);
+        } else {
+            filteredSpecies.sort(sortRevAlphabetically);
+        }
+    }
+
+    // select results to show based on page number
+    if (PAGINATION_EXISTS) {
```
kwentine, line 139: You could separate filtering and pagination by introducing a `getCurrentPage(speciesList, state)` function that could be ...you guessed it, tested separately :)

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+}
+
+
+/**
+ * Debounce a function (used for species search)
+ * @param {Function} callback - Function to debounce.
+ * @param {number} delay - Delay in ms.
+ * @returns {Function} - A debounced version of the function.
+ */
+function debounce(callback, delay) {
```
kwentine, line 224: Nice! See [Eloquent Javascript](https://eloquentjavascript.net/15_event.html#h-AOVmaqj10I) if you want a leisurely read on the topic :)

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/**
+ * Handles the search input event.
+ *
+ * @param {Event} event - The input event from the search box.
+ */
+function speciesSearch(event) {
+    state.searchText = event.target.value.toLowerCase();
+    if (PAGINATION_EXISTS) {
```
kwentine, line 241: Consider:
```javascript
state.searchText = event.target.value.toLowerCase();
state.currentPage = 1
species = filterSpecies({...state})
page = getCurrentPage(species, {...state})
renderSpeciesCards(page, root)
```

.pre-commit-config.yaml
```diff
@@ -4,6 +4,7 @@ repos:
   hooks:
   - id: check-yaml
   - id: check-json
+    exclude: ^hugo/layouts/index\.json\.json$
```
kwentine, line 7: Should be removed if we forsake the JSON

kwentine (approved):
Great work, this is in good shape :) I have now gotten to the more detailed type of review I lamented overseeing last week. It brings a lot of small comments, but I hope that you can focus on what you find interesting/relevant and discard the rest. And when you are satisfied, 🚀 🐿️

hugo/assets/js/species_cards.js
RMCrean: Thanks for the tip! I agree it is good to be consistent and I think also predictable with naming so I have updated to only have `SORT_OPTIONS` as capitalized. commit: 8f71dba424b93c7fc2dfba1621dd5ec68a081672

.pre-commit-config.yaml
RMCrean: Good spot! Now removed in commit 697bf79b82abe2978e890b8a43f0ebd73329c3f6 :)

hugo/assets/js/species_cards.js
RMCrean: Fair point, I added a Playwright test to specifically catch a change in these date formats now. Not a perfect solution perhaps but hopefully enough to make one consider. Fixing the test hopefully results in reading the test docstring which would let the dev know to change the JS too. commit 267ac4e593bd12d801d7317ecf6526227867cb04

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+ */
+function prepareSpeciesPage() {
+    const templates = document.querySelectorAll('template[id^="card-"]');
+    templates.forEach(template => {
+        const title = template.content.querySelector('.science-name').textContent.toLowerCase();
+        const subtitle = template.content.querySelector('.common-name').textContent.toLowerCase();
+        const species = {
+            id: template.id,
+            speciesText: title + ' ' + subtitle, // used for searching
+            last_updated: template.content.querySelector('.card-text').textContent.replace('Last updated: ', ''),
```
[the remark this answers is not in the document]
RMCrean: The next opportunity we have for a JSON like structure I'll try and remember to consider this one, because that does look like a nice alternative! But for now I'll stick to the <template> approach because I think that works nicely and I don't think it makes sense to do another (big) round of refactoring.

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+    numbMatches: 0 // updated when search ran.
+};
+
+
+/**
+ * Runs on page load.
+ * Gets the info needed to search and order the species cards from each <template> card.
+ * Runs the render function to display the species cards.
+ */
+function prepareSpeciesPage() {
```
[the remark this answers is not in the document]
RMCrean: Thanks, following the thought process of isolating things was really useful! I've done that now as you will soon see in my upcoming commits!

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+}
+
+const sortRevAlphabetically = (a, b) => -sortAlphabetically(a, b);
+
+
+/**
+ * Run on page load and every time an event listener is triggered.
+ * Based on current state, updates: species cards, pagination buttons and the no results card.
+ */
+function renderSpeciesCards() {
```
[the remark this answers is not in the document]
RMCrean: I have now and implemented it :) (see main comment).

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+
+
+/**
+ * Search/filter all species and order them according to active dropdown selection
+ * The id's of each species <template> is returned (in order of display).
+ *
+ * @returns {Array} - An array of filtered species HTML <template> ids.
+ *                    (Returns empty array if no matches.)
+ */
+function filterAndOrderSpecies() {
```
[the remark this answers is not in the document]
RMCrean: Yes, now it has become (each 3 params are in the state): 
`function filterAndOrderSpecies(speciesData, searchText, sortOrder)`

hugo/assets/js/species_cards.js
RMCrean: Yes! This happened naturally once I changed to limit where I update the state object. Nice to follow the single responsibility principle as much as possible too :)

hugo/assets/js/species_cards.js
RMCrean: I've now implemented something much closer to this! (see main comment)

hugo/assets/js/species_cards.js
```diff
@@ -0,0 +1,284 @@
...
+
+
+/**
+ * Runs on page load.
+ * Gets the info needed to search and order the species cards from each <template> card.
+ * Runs the render function to display the species cards.
+ */
+function prepareSpeciesPage() {
+    const templates = document.querySelectorAll('template[id^="card-"]');
+    templates.forEach(template => {
```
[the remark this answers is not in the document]
RMCrean: Thanks for the pointer, I will be more considerate about when I use `forEach` vs `map` in the future. I think in this case using `forEach` is not so bad as the function (now named `initState()` ) has the explicit purpose of modifying the state object if that makes sense?

RMCrean:
Thanks @kwentine, this has been a really fun exercise and no doubt beneficial for my coding too. 

I especially enjoyed trying to isolate the code to be more testable and learning about things like `<template>`. If times allows for it later I'll probably try and apply some of these principles to the other bits of JS code we have in the Hugo site. 

Again, I really appreciate all the pointers and tips! 

Feel free to take a look if you want (commit 037f546fdec9233c211850501e05e264e72693e1 has the bulk of the changes) but otherwise I'm happy to merge at this point :)

kwentine:
Let's merge this, thanks and congrats for great work! Really appreciate that you went the extra mile in 037f546, all this refactoring took a lot of perseverance :)  🚀 🚀 ⏮️ 📄 ⏭️

PR #76, opened 2024-12-03
Temporary fix for handling inaccessible data track url 

brinkdp:
For the next release, it would be good to have a fix in place for the in inaccessible _L. trigynum_ link: b9caa0a. If the broken link gets fixed in the foreseeable future, supporting it it would mainly be a matter of reverting the edits in b9caa0a. But if the solution is to ask the submitter to upload the file anew, there will likely be a new url that needs to be added.

The other two commits are side things that I though would be of interest to have for the next release.

The first commit regards modifying the playwright test back to a previous state where it catches and warns for the broken link. I thought this was a fun exercise in applying the playwright tests in the troubleshooting work-flow. Command:

```
pytest playwright/tests/test_species_download_page.py::test_table_links  --base-url http://localhost:8080 
```

The third commit is a mini-experiment to improve the GitHub-Zenodo integration. Last time, the release notes were not added to the Zenodo record automatically as I thought they would (I edited that manually). I think the Abstract key is the culprit, so I removed it to see if it works better next time we make a release.

Anyway, I hope this should be straight-forward to review and that it can be merged and included for the next release.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -28,7 +28,6 @@ authors:
     orcid: "https://orcid.org/0000-0001-5875-8429"
 repository-code: "https://github.com/ScilifelabDataCentre/genome-portal"
 url: "https://genomes.scilifelab.se"
-abstract: "The Swedish Reference Genome Portal is a service facilitating access and discovery of genome data of non-model eukaryotic species studied in Sweden."
 license: MIT
 identifiers:
   - description: "This is the collection of archived snapshots of all versions of the Swedish Reference Genome Portal"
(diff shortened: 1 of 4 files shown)


RMCrean (approved):
Thanks for this :) 

I tested running:

`./scripts/dockermake SPECIES=linum_tenue`

Both with my currently populated local `data `folder and then after deleting my local `data `folder. Both occasions worked well so LGTM. Before making a new release I can also run `make` on the dev cluster as 1 extra layer of confidence.

brinkdp:
Thanks, sounds like plan!

> ./scripts/dockermake SPECIES=linum_tenue

Small remark: this fix is for `linum_trigynum` not `linum_tenue` 😄 But if you re-initiated your data whole dir and not just `data/linum_tenue` after deleting it, it would have tested for _L. trigunum_ too.

kwentine (approved):
🚀 🌱

PR #78, opened 2024-12-09
EB-396: Generate assembly aliases as part of the data build

kwentine:
# Context
Alias files for ENA assemblies are currently being generated by manually invoking `get_aliases_from_ENA_fasta.py`. This presents two disadvantages:
- It adds a manual step to our workflow
- The aliases are made available to JBrowse by hosting them at a conveniently accessible location. Since neither Google Drive nor Nextcloud proved suitable, we fallback to `scripts/data_stewardship_alias_files_temp_storage` in this repository. This forces us to check in sometimes large files to version control unnecessarily.

# Contribution
This PR introduces a `make aliases` recipe that will output one `aliases.txt` per species. Thus, `make build` will store the alias file alongside other genomic assets to be deployed on k8s.

# Implementation
- The logic in `get_aliases_from_ENA_fasta.py` was ported to a `sed` script in `scripts/aliases`. This is a brittle move perhaps, so I am open to keeping the logic in Python. This solution got me started quickly however, since the Docker image we use for the build does not currently have Python installed. And I learned how to branch in `sed` which was kinda 😎 
- `make aliases` generates one `aliases.txt` per species. If the species has multiple assemblies,  all the aliases will be in this file. It is slightly inefficient since `aliases.txt` will be empty for non-ENA assemblies, but I thought the gained simplicity was worth it.

- If an assembly's `alias` field is omitted in `config.yml`, it defaults to `aliases.txt` (see 0bc26e)

# Testing
- Unit test are found in `tests/bats/aliases` but running them requires some setup that I can help you with.
- Otherwise, for example, at the root of the repository:
```sh
# Optionally, you can add our scripts directory to your executable search path
PATH=$(pwd)/scripts:${PATH}
# After setting SPECIES in the environment, all future make commands will operate on it 
export SPECIES=linume_tenue
dockermake aliases
dockermake jbrowe-config
```


RMCrean (commented):
Impressive work! 

> The logic in get_aliases_from_ENA_fasta.py was ported to a sed script in scripts/aliases. This is a brittle move perhaps, so I am open to keeping the logic in Python. This solution got me started quickly however, since the Docker image we use for the build does not currently have Python installed. And I learned how to branch in sed which was kinda :sunglasses:

Impressive scripting. I think if we see down the line a need for quite a bit more scripting it could be worth considering adding Python to the image to make it more flexible. I think this is something you and Daniel should decide though as you both know this part of the repo much better than me.  


> make aliases generates one aliases.txt per species. If the species has multiple assemblies, all the aliases will be in this file. It is slightly inefficient since aliases.txt will be empty for non-ENA assemblies, but I thought the gained simplicity was worth it.

I think having an empty alias.txt file for species without an ENA file is not the end of the world and a solution that scales fine when we consider how many species we expect to end up with in the end. 

Everything looks okay to me. However, when I deleted my `data` dir and ran:

```
./scripts/dockerbuild data
SWG_TAG=local ./scripts/dockermake
```
I got the following error (it repeats many times):
```
# rest of output before all fine
Indexing data/meganyctiphanes_norvegica/annotations.protein_coding_gene_models.non_redundant_added_stop_codons.gff.bgz
[E::get_intv] Failed to parse TBX_GENERIC, was wrong -p [type] used?
The offending line was: ""
# error message above repeated lots (thousands of times)
# output continues as normal afterwards... 
Indexing data/linum_trigynum/Ltrigynum_v1_genes.gff.bgz
Indexing data/skeletonema_marinoi/GCA_030871285.1_GU_Smar_1.1.8_genomic.gff.bgz
```
When I open meganyctiphanes_norvegica in the browser page afterwards both the protein coding genes and ref genes seem to show up fine:
Screenshot from 2024-12-10 13-47-13


Could this be related? I don't know if either of you are able to reproduce the error?

brinkdp (approved):
I agree with Rory: this is impressive work! Thank you for the excellent write-up and the clear test examples.

I too want to say well done on converting the python script to sed! Funny story is that that script actually started as a sed one-liner, so it is fun to see it going full circle! The main reason I reimplemented it in python was to be able to add more modules to handle edge-cases down the line; however, that has not yet surfaced due to other prioritizations :smile:  I was not aware that it was possible to do branching within a `sed` command - I would just have wrapped it in a conditional block. I assume that it is more performance efficient to branch like this too. It was very cool to see, thanks for investigating this!

For testing, I ran the following and everything worked smoothly:

```
SWG_TAG=local ./scripts/dockermake aliases SPECIES=linum_tenue
diff data/linum_tenue/aliases.txt ~/Downloads/camgyj01.fna.alias
SWG_TAG=local ./scripts/dockermake jbrowse-config
SWG_TAG=local ./scripts/browse data/linum_tenue
```

New alias looks good and tracks are displayed as intended in JBrowse. Very good!

I think this is a brilliant addition to the build process. I like that the option to specify alias url is still an option, since that will be useful for the inevitable edge-cases. The Clouded Apollo mito assembly is one of those: the required header synonym for the annotation track is not part of the ENA header substring and thus had to be manually added to the alias.

I did also test with the multi-assembly config from #77 (with omitted alias urls), despite it being an edge-case. All aliases were written to the same file as intended, and the `dockermake jbrowse-config`put the aliases in the correct places in the final `config.json`. Great idea to handle it thusly, @kwentine. 

For a sanity-test, I added back the url for the mito alias to the multi-assembly config. As expected, it generated `alias.txt` for the primary genome assembly, and loaded the mito alias from the GitHub location. This gave a fully functional JBrowse multi-assembly session. I am very happy with this functionality for the time being. The drawback is that we still would need to store these edge-case alias files on GitHub for now, but as we learn more about such needs, I think we can continue to add that functionality to the `aliases` recipe, or even to the python script.

I like how this will store the alias files on the permanent volume. If we want to update an alias file, for instance if new tracks are added and more header variants need to be added to the alias, will `make` understand that the file needs to be updated, or do we need handle that manually? 


@RMCrean: 

>I think if we see down the line a need for quite a bit more scripting it could be worth considering adding Python to the image to make it more flexible.

Sounds good to me. For the aliases, this implementation seem to cover a lot of ground.
But for the upcoming slew of helper scripts that I have started working on with Angela, I think they will be mainly in python. I am very keen to learn more about docker, and it would be a fun exercise for me to set up a docker file for the dependencies of the data stewardship scripts. Are there any reasons for not including python in the docker images, apart for it not being used in the data builder processes that we are talking about here?

>I think having an empty alias.txt file for species without an ENA file is not the end of the world and a solution that scales fine when we consider how many species we expect to end up with in the end.

Agreed, especially since it will not affect the JBrowse session. As long as `alias.txt` is present in the path that is specified in `config.json`, JBrowse will not complain. In this case, we can use it to our benefit to avoid having to consider the consequences of an empty alias file. (Likewise, this behavior is the cause of "the empty track window" error when an alias file has been correctly found and loaded, but that its contents are incorrect for the current assembly and tracks.)

>Indexing data/meganyctiphanes_norvegica/>annotations.protein_coding_gene_models.non_redundant_added_stop_codons.gff.bgz
>[E::get_intv] Failed to parse TBX_GENERIC, was wrong -p [type] used?

Hmm, that looks like a tabix error. I did not delete my data directory for this test, so I did not encounter this, but I have seen similar errors before. Since it relates to the krill, is this the step that take ca 20 min to run? Or did you get the error early on in the build process run? 

I wonder why it shows up in JBrowse without much error 🤔  Did the processes you ran include the copying of data to the hugo directory or are you in fact looking at an older compilation of the data? Or does the error imply that the tabix process just skipped the problematic rows? (this is not how I remember tabix operating)

I can take a look at replicating this error when I have time, if you want.

kwentine:
> I think if we see down the line a need for quite a bit more scripting it could be worth considering adding Python to the image to make it more flexible.

I completely agree, this is a tipping point. Should we need, for example, to support aliasing another flavor of FASTA headers, then I would switch to Python.  

> I think having an empty alias.txt file for species without an ENA file is not the end of the world

Agreed. The more annoying inefficiency is that the mere streaming of a large FASTA through `sed`, even when no substitutions are performed, takes quite a lot of time. Again, not the end of the world.


> I got the following error (it repeats many times):
> ```
> Indexing data/meganyctiphanes_norvegica/annotations.protein_coding_gene_models.non_redundant_added_stop_codons.gff.bgz
> [E::get_intv] Failed to parse TBX_GENERIC, was wrong -p [type] used?
> The offending line was: ""
> ```

Thanks for surfacing this. Those are in fact warnings emitted by `tabix` that cannot quite stomach the empty lines found in the Krill GFF, but produces a valid index nonetheless . This originally motivated [EB-378](https://scilifelab.atlassian.net/browse/EB-378?atlOrigin=eyJpIjoiNzc3N2FmMmVlOTRhNGU5ODg2NjY4NWUyZTViNjAxYWEiLCJwIjoiaiJ9) and I could update the issue description thanks to your detailed log trace.

RMCrean:
Good to hear you're both aware of and on top of this issue. No need to try and "_fix_" it in this PR then. 

@brinkdp 
> Sounds good to me. For the aliases, this implementation seem to cover a lot of ground.
But for the upcoming slew of helper scripts that I have started working on with Angela, I think they will be mainly in python. I am very keen to learn more about docker, and it would be a fun exercise for me to set up a docker file for the dependencies of the data stewardship scripts. Are there any reasons for not including python in the docker images, apart for it not being used in the data builder processes that we are talking about here?

Nothing wrong with having a python docker image for the helper scripts and as you say this could/should be a separate docker image and not shared with the data builder image. By all means create a new dockerfile if you would like to :). P.S. If you plan to have a `requirements.txt `file for all the dependencies of these scripts in this case you could name it something like `requirements-[distinguishing-name].txt` (e.g. [see how they do it here](https://github.com/fastapi/fastapi) ).

brinkdp:
> can we assume that `contig: ` would either never or always be present in the headers ?

Indeed, this can not be assumed. From my anecdotal experience, `contig:` is present in most of the assemblies that we have encountered, but not all. I could hypothesize that the assemblies that have been produced and brokered with help from SciLifeLab staff all the substring, but I do not know for certain.  

>To manually handle alias patches, I have a few ideas that we could discuss. One of them could be an approach akin to how config.json is handled: we can put a default aliases.txt in config that would be extended by whatever can be parsed from FASTA headers. In fact, going that way, we could move the Clouded Apollo aliases to config/parnassius_mnemosyne/aliases.txt. Do you see what I mean ?

I like this idea! Adopting this philosophy also to the aliases would be good for consistency in the work-flow.

kwentine:
> I like this idea! Adopting this philosophy also to the aliases would be good for consistency in the work-flow.

I have started an implementation, but will keep it for a separate PR. Stay tuned :)

I'm merging this now after having cleaned up the history a wee bit  🚀

PR #79, opened 2024-12-16
Revert "EB-405: Add mitochondrial assembly to Clouded Apollo"

RMCrean:
Reverts ScilifelabDataCentre/genome-portal#77


RMCrean:
Revert accidental merge conflict resolution that set files back to their previous state

PR #81, opened 2024-12-20
EB-134: Add a Content Security Policy to the website

RMCrean:
This PR is about adding a [content security policy](https://web.dev/articles/csp) to the website. 

This essentially involves removing inline JS and creating a whitelist inside the `docker/nginx-custom.conf` for all the external connections we expect to make.

The genome browser page posed some extra challenges when creating this:

1. In order to remove `'unsafe-inline'` from  the `script-src` line required updating each species `config.json` to disable [google analytics](https://jbrowse.org/jb2/docs/config_guides/disable_analytics/). Just whitelisting google analytics webpages did not work. 
2. Trying to remove `'unsafe-inline'` from  `style-src` was unsuccessful because JBrowse relies on it. I think it is okay to leave `'unsafe-inline'` in `style-src`and we can be happy we got the rest. 

@brinkdp  and I already spoke and he is working on including the disable google analytics param in the `config.json` for any future species. (If the param is not included the browser page still works as expected, you just see some errors if you open up the console).

### DIFF
--- a/docker/nginx-custom.conf
+++ b/docker/nginx-custom.conf
@@ -20,7 +20,15 @@ server {
     add_header X-Frame-Options "SAMEORIGIN"; # don't allow iframing
     add_header X-Content-Type-Options "nosniff"; # prevent MIME type sniffing
     add_header X-XSS-Protection "1; mode=block"; # Cross-site scripting (XSS) filter
-
+    add_header Content-Security-Policy "
+        default-src 'self'; 
+        script-src 'self' https://cdn.jsdelivr.net http://matomo.dc.scilifelab.se https://www.google.com https://www.gstatic.com/ https://cdn.datatables.net https://code.jquery.com https://unpkg.com;  
+        style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net https://cdn.datatables.net https://unpkg.com; 
+        font-src 'self' https://fonts.gstatic.com https://cdn.jsdelivr.net; 
+        img-src 'self' data: https://tile.gbif.org https://api.gbif.org;
+        connect-src 'self' http://matomo.dc.scilifelab.se/ https://raw.githubusercontent.com https://analytics.jbrowse.org; 
+        frame-src 'self' https://www.google.com https://matomo.dc.scilifelab.se;
+    ";
     # cache media and font files for 30 days 
     location ~* \.(webp|svg|jpg|jpeg|png|gif|woff|woff2|ttf|eot|otf)$ {
         add_header Cache-Control "public, max-age=2592000";
--- a/docker/nginx-custom.conf
+++ b/docker/nginx-custom.conf
@@ -20,15 +20,7 @@ server {
     add_header X-Frame-Options "SAMEORIGIN"; # don't allow iframing
     add_header X-Content-Type-Options "nosniff"; # prevent MIME type sniffing
     add_header X-XSS-Protection "1; mode=block"; # Cross-site scripting (XSS) filter
-    add_header Content-Security-Policy "
-        default-src 'self'; 
-        script-src 'self' https://cdn.jsdelivr.net http://matomo.dc.scilifelab.se https://www.google.com https://www.gstatic.com/ https://cdn.datatables.net https://code.jquery.com https://unpkg.com;  
-        style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net https://cdn.datatables.net https://unpkg.com; 
-        font-src 'self' https://fonts.gstatic.com https://cdn.jsdelivr.net; 
-        img-src 'self' data: https://tile.gbif.org https://api.gbif.org;
-        connect-src 'self' http://matomo.dc.scilifelab.se/ https://raw.githubusercontent.com https://analytics.jbrowse.org; 
-        frame-src 'self' https://www.google.com https://matomo.dc.scilifelab.se;
-    ";
+
     # cache media and font files for 30 days 
     location ~* \.(webp|svg|jpg|jpeg|png|gif|woff|woff2|ttf|eot|otf)$ {
         add_header Cache-Control "public, max-age=2592000";
--- a/hugo/assets/js/recaptcha.js
+++ b/hugo/assets/js/recaptcha.js
@@ -1,8 +1,0 @@
-/*
-Google reCAPTCHA v2 callback
-*/
-var onloadCallback = function () {
-    grecaptcha.render(id = "recaptcha-widget", {
-        'sitekey': '6LfLRQMqAAAAACSectjtigWkNvif1G2J8EFlj1nV'
-    });
-};
--- a/hugo/assets/js/recaptcha.js
+++ b/hugo/assets/js/recaptcha.js
@@ -1,0 +1,8 @@
+/*
+Google reCAPTCHA v2 callback
+*/
+var onloadCallback = function () {
+    grecaptcha.render(id = "recaptcha-widget", {
+        'sitekey': '6LfLRQMqAAAAACSectjtigWkNvif1G2J8EFlj1nV'
+    });
+};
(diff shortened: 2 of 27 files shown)


docker/nginx-custom.conf
```diff
@@ -20,7 +20,15 @@ server {
...
     add_header X-Content-Type-Options "nosniff"; # prevent MIME type sniffing
     add_header X-XSS-Protection "1; mode=block"; # Cross-site scripting (XSS) filter
-
+    add_header Content-Security-Policy "
+        default-src 'self'; 
+        script-src 'self' https://cdn.jsdelivr.net http://matomo.dc.scilifelab.se https://www.google.com https://www.gstatic.com/ https://cdn.datatables.net https://code.jquery.com https://unpkg.com;  
+        style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://cdn.jsdelivr.net https://cdn.datatables.net https://unpkg.com; 
+        font-src 'self' https://fonts.gstatic.com https://cdn.jsdelivr.net; 
+        img-src 'self' data: https://tile.gbif.org https://api.gbif.org;
+        connect-src 'self' http://matomo.dc.scilifelab.se/ https://raw.githubusercontent.com https://analytics.jbrowse.org;
```
kwentine, line 29: Is there a reason you opted for the trailing `/` syntax here, and the `http` scheme 🤔

hugo/assets/js/recaptcha.js
```diff
@@ -0,0 +1,8 @@
+/*
+Google reCAPTCHA v2 callback
+*/
+var onloadCallback = function () {
```
kwentine, line 4: Why not opt for `document.addEventListener` here as in `init_datatables.js` (for example) ? I was puzzled, by what magic is the event listener registered ? Is it a convention used by Google reCAPTCHA ?

hugo/layouts/contact/list.html
```diff
@@ -8,9 +8,16 @@
 
 {{ define "script_includes" }}
 
-<!-- Load the Google reCAPTCHA API asynchronously and defer its execution until the page has finished parsing -->
+<!-- onload callback function needs to be loaded before reCAPTCHA API to avoid race conditions. -->
```
kwentine, line 11: This comment is a bit cryptic to me, perhaps related to my comment above regarding `onLoadCallback`

hugo/layouts/partials/footer.html
```diff
@@ -79,17 +79,5 @@ <h2 class="scilife-footer-titles mt-2 mb-1">Stay Updated</h2>
...
-    _paq.push(['enableLinkTracking']);
-    (function () {
-        var u = "//matomo.dc.scilifelab.se/";
-        _paq.push(['setTrackerUrl', u + 'matomo.php']);
-        _paq.push(['setSiteId', '11']);
-        var d = document, g = d.createElement('script'), s = d.getElementsByTagName('script')[0];
-        g.async = true; g.src = u + 'matomo.js'; s.parentNode.insertBefore(g, s);
-    })();
-</script>
+{{ $matomoJS := resources.Get "js/matomo.js" | fingerprint | minify }}
```
kwentine, line 82: A Hugo idiom I like for all these cases:
```
{{ with resources.Get "js/matomo.js" | fingerprint | minify }}
  <script src="{{ .RelPermalink }}"></script>
{{ else }}
  {/* Error handling */}
{{ end }}
```

kwentine (commented):
Thanks for this thorough work. It was interesting to learn about CSP using the linked resources.

How did the idea to enforce a CSP come from originally ? I am divided: while the extraction of inline code is a valuable refactoring in itself, I wonder if maintaining a list of directives in `nginx,conf` is worth it. Quoting [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html):

> Even on a fully static website, which does not accept any user input, a CSP can be used to enforce the use of [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). This can help prevent malicious code from being loaded on the website if one of the third-party sites hosting JavaScript files (such as analytics scripts) is compromised.

It seems to imply that the main benefit of CSP for a static site like ours would be to check the integrity of scripts, rather than mitigating unlikely XSS threat by restricting subresource origins. But we actually omit the former. 

I don't strongly object merging this though, but some documentation snippet should perhaps be included to avoid some head scratching the day one of us includes a new third-party script, everything working well locally, buy breaks when deployed because of a stale `nginx.conf` 😅

hugo/layouts/partials/footer.html
RMCrean: If the file doesn't exist the build will fail because of the fingerprint step, so the above `{{ else }}` block would never be triggered. We could do something like this I suppose:

```
{{ $jsFile := "js/does_not_exist.js" }}
{{ $resource := resources.Get $jsFile }}
{{ if $resource }}
  {{ $minified := $resource | fingerprint | minify }}
  <script src="{{ $minified.RelPermalink }}"></script>
{{ else }}
  <script>
    console.error("Failed to load Matomo tracking script: {{ $jsFile }}.");
  </script>
{{ end }}
```
But perhaps it is better to let the build fail and fix the file path?

hugo/assets/js/recaptcha.js
RMCrean: Yes, I used it because it was what Google recommended in their docs: https://developers.google.com/recaptcha/docs/display. My understanding is that the `onloadCallback` is called by this line in `hugo/layouts/contact/list.html`:

```
<script src="https://www.google.com/recaptcha/api.js?onload=onloadCallback&render=explicit" async defer></script>
```

hugo/layouts/contact/list.html
RMCrean: I tried to summarize the contents of their docs page (https://developers.google.com/recaptcha/docs/display) in one line but I agree, that was not easy to read. I've now added a bit more documentation and added a link to their docs page inside the file which will hopefully help? commit dbc01b7f58fb15189a3b291c9e6486372c69e85c

docker/nginx-custom.conf
RMCrean: Good spot thank you, that was a mistake on my end. I'm about to write in the main comment that I agree with your line of thinking and that we can skip the CSP but just keep the other changes then. So I will not commit a fix as this line will be removed anyway.

RMCrean:
@kwentine, Thanks I think you're right. We now have SRI checks on the external JS scripts anyway and I agree with you, the inclusion of a CSP does add more complexity when it would come to adding new functionality/3rd party scripts, so as it doesn't bring any extra benefits I'm also in favour of removing it. I have therefore now removed the CSP from `docker/nginx-custom.conf` (commit 9e4fb6a2bd92f3896f6fe00efa73829a8e3cf61b) but kept the rest of changes. 

Please let me know if you're happy with the other comments/changes. Thanks again!

hugo/assets/js/recaptcha.js
kwentine: Okay I understand thanks! Out of curiosity: why did you prefer explicit rendering to the [automatic](https://developers.google.com/recaptcha/docs/display#auto_render) markup-driven rendering ? It could be one script less to load, if we don't plan a more advanced use that would warrant an `onloadCallback` for custom logic.

kwentine (approved):
👍🏻 👍🏻

hugo/assets/js/recaptcha.js
RMCrean: Honestly can't remember why/if I even considered that option when I implemented it (maybe I skimmed the docs too much and saw "explicit rendering" and thought - that's what I want, I'll just look at this part) but your suggested approach is much simpler I agree, so I've now made that change with commit 1234a7eac9693caa250e16a91a6a591cef8dde09. Thanks!

hugo/layouts/partials/footer.html
RMCrean: I see your point, in this case I suppose I lean more towards not adding the custom error handling because I think the default error you would get is sufficient to explain the problem. But it doesn't matter now as your other suggestion meant that all this code is no longer needed haha.

PR #84, opened 2025-01-17
EB:203 UI consensus

RMCrean:
This PR covers [EB:203](https://scilifelab.atlassian.net/browse/EB-203) which is a combination of many tasks around the UI for the portal. Please feel free to explore the changes locally and let me know what you think or if you spot any issues. **Of course, happy to make changes/additions both in the code or more stylistically**, I see this as a good way to get the ball rolling. 

**Tasks (with specific comments associated):** 

#### 1. Use a darker border of the tabs of the species pages 
Reason: some projectors struggle displaying light grey, perhaps use medium grey? Or consider another style for displaying the tabs, so there is more colour contrast. 
Done.

#### 2. Website logo is too long for a mobile phone screen, fixable?
Extra media queries added to handle this for all cases except a small phone.  For a small phone screen reducing the logo size makes it too small, so this I think is a good compromise and improvement on what we had. 

#### 3.  Decide whether "first date on portal" should be kept in the Downloads table
So far kept, don't have any strong opinions on this though. 

#### 4. The website logo cannot be seen in a mobile or tablet using “dark mode"
A "Dark mode" theme has now been added to the genome portal. Dark mode will be automatically activated for those users that have that as there default preference. It can also be modified in the navbar. 

Screenshot from 2025-01-17 13-36-29

A side note here is that JBrowse also has light and dark themes (tools -> preferences -> themes). I have left that alone in this PR but it is something we can consider for the future. 

#### 5. Suggested format for footer part of software versions: "Website code on Github (v1.0.0)". Perhaps release date is not important to show, as for citation what matter is the date of visit/use
I have edited the footer to remove the date but wrote the following instead: "Swedish Reference Genome Portal version: v1.1.0" 
Happy to change it though. 

##### 6.  See if copy to clipboard can copy formatted text not plain text 
It's possible to copy in a non-plain text way like HTML but the problem is this is not reliably transferable as hoped when you then paste it. For example copying and then pasting into google docs gives text like: `<em>this text should be italics but instead the pasted text is just HTML <em>`, instead of directly converting it to italicized text. Essentially, there doesn't seem to be a good solution for this that will reliably paste how we want it to. So I think sticking with plain text copying seems to be the best option.

##### 7. Increase font size for the “title“ of each of the questions in the FAQ page (taken from EB-402)
Done, can consider further modification if desired. Such as using [bootstrap's accordion component](https://getbootstrap.com/docs/5.3/components/accordion/) for each faq question etc...


brinkdp (approved):
Thanks for this neat collection of updates. Aside from a tiny CSS issue for the dark mode (see 4 below), LGTM! 

**3. First date on portal**
Can't say that I have strong opinions on this either, and there is redundancy between the change-log and first date on portal. What speaks towards keeping it in the table for me is that it makes the metadata richer, which is something we should strive for (FAIR principle F2).

**4. Dark mode**
Great job on this! I will use this mode a lot. 
I went through all pages and the only thing I could spot was that three h3 elements in the Contribute page were still rendered in black. The black font colour is hard-coded in `.scilife-contrib-subheader` in `styles.css`. I don't think this class is used anywhere else on the site, so I suspect that the line can just be dropped?

Less important, and very subjectively, I was also wondering about the table headings (glossary, species download tabs etc.). I find that they appear very bright in dark mode. Could it be worthwhile to set them to a darker gray?

**5. Footer**
I think the new version adds clarity and is a good choice. What was the reason to remove the date? Is is because the other DC portals do not display it?

(I understand that a user can click on the version number and find the deployment date on GH, so the information is still there, indirectly. I was just thinking that if we, for some reason, want the want users to see the last updated date, it is more clear to display it explicitly.)

**6. Copy-to-clipboard**
Agreed, let's keep it as plain text.

hugo/assets/js/determine_theme.js
```diff
@@ -0,0 +1,12 @@
+/*
+This script runs as early as possible to detect if a user prefers dark or light mode.
```
kwentine, line 2: Great docstrings 👍🏻

hugo/assets/js/determine_theme.js
```diff
@@ -0,0 +1,12 @@
...
+Running as early as possible prevents the "white flash" (if user prefers dark mode) seen when loading the page otherwise.
+This script works with "change_theme.js" which provides a way to swap the mode/theme once the page is loaded.
+
+Note: extra brackets on the function make it a: Immediately Invoked Function Expression
+*/
+(function() {
+    const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+    const currentTheme = localStorage.getItem('bsTheme') || (prefersDark ? 'dark' : 'light');
```
kwentine, line 10: Since `window.matchMedia` is called systematically, is there really a need for the local storage 🤔 I would understand the purpose if `matchMedia` were a costly operation, but then the cache should get checked first.

hugo/assets/css/styles.css
```diff
@@ -135,8 +126,7 @@ body {
     margin-right: 30%;
   }
 }
-
-@media (min-width: 992px) and (max-width: 1199.98px) {
+@media (min-width: 992px) and (max-width: 1199px) {
```
kwentine, line 129: I barely know a thing about the topic, so out of curiosity: how do you come up with the relevant breakpoints like 576, 768, 992 etc. ?

kwentine (approved):
I contribute my enthusiastic approval to the sought consensus! I appreciate the dark mode and switching mechanism, as well as the peek into its implementation techniques that the review afforded.

hugo/assets/css/styles.css
RMCrean: I suppose you can pick any number but these numbers are related to [what bootstrap use, so I try to use them mostly](https://getbootstrap.com/docs/5.3/layout/breakpoints/). 

One nice related thing with with these breakpoints and bootstrap is that all (or almost all - not sure) bootstrap classes can be made sensitive to specific breakpoints: 

E.g.:
`<p class="text-start text-lg-center">  blah blah blah </p>`

Means the text will be left-aligned (text-start) on smaller screens but in the center for screens with size larger than "lg" breakpoint.

hugo/assets/js/determine_theme.js
RMCrean: Just to make sure I'm on the same page, when you said "cache" you mean localStorage right? Or do you mean some another kind of cache/store? 

I think the local storage should be used to preserve a users preference (if they manually changed the theme themselves for example) if they come back to the site another time. But, as you say then the cache should be checked first so it could be something like below instead: 

```js
(function() { // # noqa e404 - quentin 1 liner ;)  
    const storedTheme = localStorage.getItem('bsTheme');
    if (storedTheme) {
        document.documentElement.setAttribute('data-bs-theme', storedTheme);
    } else { 
        const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
        const currentTheme = prefersDark ? 'dark' : 'light';
        document.documentElement.setAttribute('data-bs-theme', currentTheme);
    }
})();
```
Does that sound good?

hugo/assets/js/change_theme.js
```diff
@@ -0,0 +1,50 @@
...
+
+The html associated with this JS is in a Hugo partial "change_theme.html", which is used in the navbar.
+*/
+document.addEventListener('DOMContentLoaded', () => {
+    const htmlElement = document.documentElement;
+    const themeOptions = document.querySelectorAll('.dropdown-item[data-bs-theme-value]');
+
+    // helper function to swap image srcs.
+    const updateImages = (theme) => {
+        document.querySelectorAll('img[data-dark-src]').forEach((img) => {
```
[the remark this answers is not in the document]
RMCrean: Thanks, I've added that change in now in commit 5441fbf01c0190045176bf70acc26b552f317f33

hugo/assets/js/determine_theme.js
RMCrean: No problem at all. That sounds good. Haha, yes it was designed to prevent (or perhaps encourage ;) ) something like that). I change the code inspired by your 1 liner to a _relatively_ short code block which now adds it to localStorage, commit e8e85d2fa13486108983cfb767835298140898ea :).

apfuentes:
Hi @RMCrean,

I'm impressed by the really cool features you implemented in the UI! 

A few minor comments quoted below:

> #### 1. Use a darker border of the tabs of the species pages
It looks great!

> #### 2. Website logo is too long for a mobile phone screen, fixable?
> Extra media queries added to handle this for all cases except a small phone. For a small phone screen reducing the logo size makes it too small, so this I think is a good compromise and improvement on what we had.

OK, let's leave it as it is. Either way, we don't expect researchers will use their mobile phone that much to examine their genomic data.
 
> #### 3. Decide whether "first date on portal" should be kept in the Downloads table
> So far kept, don't have any strong opinions on this though.

No strong opinions on this either, just that I wondered about this when evaluating which text sections could be excluded to reduce information load of users. But as @brinkdp suggests, perhaps better to keep it as part of metadata.

> #### 4. The website logo cannot be seen in a mobile or tablet using “dark mode"
> A "Dark mode" theme has now been added to the genome portal. Dark mode will be automatically activated for those users that have that as there default preference. It can also be modified in the navbar.

It looks amazing! Thanks for adding this feature :)

> A side note here is that JBrowse also has light and dark themes (tools -> preferences -> themes). I have left that alone in this PR but it is something we can consider for the future.
Sure, good idea, but not priority at the moment.

> #### 5. Suggested format for footer part of software versions: "Website code on Github (v1.0.0)". Perhaps release date is not important to show, as for citation what matters is the date of visit/use
> I have edited the footer to remove the date but wrote the following instead: "Swedish Reference Genome Portal version: v1.1.0" Happy to change it though.

LGTM.

> ##### 6. See if copy to clipboard can copy formatted text not plain text
> It's possible to copy in a non-plain text way like HTML but the problem is this is not reliably transferable as hoped when you then paste it. For example copying and then pasting into google docs gives text like: `\<em\>this text should be italics but instead the pasted text is just HTML \<em\>`, instead of directly converting it to italicized text. Essentially, there doesn't seem to be a good solution for this that will reliably paste how we want it to. So I think sticking with plain text copying seems to be the best option.

Thanks for investigating this, plain text should suffice.

> ##### 7. Increase font size for the “title“ of each of the questions in the FAQ page (taken from EB-402)
> Done, can consider further modification if desired. Such as using [bootstrap's accordion component](https://getbootstrap.com/docs/5.3/components/accordion/) for each faq question etc...

Thanks for increasing the font size. The text is more readable now, but I must confess that I really like the look of the proposed accordion component. If it doesn't take too long, would it be possible to implement this feature in this PR? Otherwise, this could go in a separate PR.

:)

RMCrean:
@apfuentes, thanks for going through each item and checking out the changes, glad you liked them! 

I have commited the accordion version of the FAQs now if you want to take a look :)

RMCrean:
Thanks @apfuentes for going through and checking. You're right there was a difference which I have now fixed in commit 700bcfa7a6dc1bf78cd7b37425236d8538c91f9e , with the reason for the difference and fix included in the commit message. 

I will merge this PR shortly and we can discuss making a new release for this at the scrum meeting later today :)

PR #82, opened 2025-01-07
EB-410: Adding custom aliases to the data build

kwentine:
Support custom aliases in the species `config` directory, much like we support a custom config.json: an initial config/<species>/aliases.txt that is extended by aliases automatically parsed from the FASTA headers, and placed alongside other data files by the build process.

PR #85, opened 2025-01-23
EB-317 Add Amphiura filiformis to genome portal

brinkdp:
This PR is about adding the brittle star *Amphiura filiformis* to the Genome Portal. The ENA FASTA header in this assembly is an exception from the pattern that we have previously encountered, so the alias generated by the makefile will only yield an empty track. I used awk to create a custom refNameAlias, located at `config/amphiura_filiformis/aliases.txt` in anticipation of the future merge of #82.  For now, there is also a copy of the alias in temp storage so that species can be merged independently of PR82.

There is one annotation track at the moment, but more are expected to come in the future. Could you please have a look at content on the hugo pages too, to ensure that everything looks alright. 

Commands:

```
./scripts/dockermake SPECIES=amphiura_filiformis
./scripts/dockerserve
```


hugo/content/species/amphiura_filiformis/_index.md
```diff
@@ -0,0 +1,56 @@
...
+# These 3 params define the initial view for the map
+# latitude and longitude are for the map center.
+latitude: 45
+longitude: 10
+initialZoom: 2
+---
+
+### Description
+
+*Amphiura filiformis* is a burrowing brittlestar with a disc up to 10 mm in diameter with long arms (10 x disc diameter in average) that extend into the water column for suspension feeding, making it an ecosystem engineer and important food item for flatfish, cod and crayfish. It is the dominant species on many sublittoral soft bottoms down to 200 m depth in the North Sea and the Mediterranean.  *A. filiformis* has separate sexes and reproduce annually during the summer months (July-September). Adults are at least 3 years old before they are mature and have a life span of 10-20 years.  Larvae is planktotrophic and metamorphose a month later under culture conditions. Both larvae and adults can be cultured.
```
RMCrean, line 34:
```suggestion
*Amphiura filiformis* is a burrowing brittlestar with a disc up to 10 mm in diameter with long arms (10 x their disc diameter on average) that extend into the water column for suspension feeding, making it an ecosystem engineer and important food item for flatfish, cod and crayfish. It is the dominant species on many sublittoral soft bottoms down to 200 m depth in the North Sea and the Mediterranean.  *A. filiformis* has separate sexes and reproduce annually during the summer months (July-September). Adults are at least 3 years old before they are mature and have a life span of 10-20 years.  Their larvae are planktotrophic and metamorphose a month later under culture conditions. Both larvae and adults can be cultured.
```

hugo/assets/amphiura_filiformis/data_tracks.json
```diff
@@ -0,0 +1,30 @@
+[
+    {
+      "dataTrackName": "Genome",
+      "description": "Reference genome sequence",
+      "links": [
+        {"Download": "ftp://ftp.ebi.ac.uk/pub/databases/ena/wgs/public/jaz/JAZBNO01.fasta.gz"},
```
RMCrean, line 6:
```suggestion
        {"Download": "https://ftp.ebi.ac.uk/pub/databases/ena/wgs/public/jaz/JAZBNO01.fasta.gz"},
```

hugo/content/species/amphiura_filiformis/_index.md
```diff
@@ -0,0 +1,56 @@
...
+
+*Amphiura filiformis* has exceptional adult regenerative abilities and can regenerate fully functional arms in a matter of weeks, making this species an emerging model for regeneration and stem cell biology in biomedical research. Major scientific targets include cellular and molecular mechanisms underlying regeneration with important links to neuroscience, stem cell biology and neuropeptide structure and function in the absence of a centralised nervous system.
+
+Echinoderms have the potential to offer viable and tractable models for molecular and cellular research on stem cells and regeneration, aided by the fact that echinoderms are characterised by a simple genomic structure. Genetic information between this species is highly conserved, compared to humans, but the nature of the common molecular regulatory pathways that facilitate regeneration has been unclear until now.
```
RMCrean, line 38: First time the term Echinoderms is mentioned, so maybe worth doing something like:
```suggestion
Echinoderms such as *Amphiura filiformis* have the potential to offer viable and tractable models for molecular and cellular research on stem cells and regeneration, aided by the fact that echinoderms are characterised by a simple genomic structure. Genetic information between this species is highly conserved, compared to humans, but the nature of the common molecular regulatory pathways that facilitate regeneration has been unclear until now.
```

RMCrean (approved):
Really nice work. And again, cool picture :) 

Two minor text suggestions and one fix for a download link (I think Hugo didn't like the use of the ftp protocol but using https instead worked fine for me at least (relates to this https://discourse.gohugo.io/t/zgotmplz/35166)

**Regarding the playwright test failure in the GH action:** 
`The heading: References is not visible on page: http://localhost:8080/amphiura_filiformis/`

If it isn't a required header for a species intro page are you okay to remove it from the 
`REQUIRED_HEADERS `list inside `playwright/tests/test_species_intro_page.py`?

brinkdp:
Thanks, I like the suggestions!

> I think Hugo didn't like the use of the ftp protocol but using https instead worked fine for me at least 

Great that you spotted this. Indeed, I normally change it to https, but here I had forgotten.

> `The heading: References is not visible on page: http://localhost:8080/amphiura_filiformis/`

In the original commit, I made the call to drop the References section since there was only one reference, which was already mentioned in the citation box. Thus, I disregarded the playwright error. But your comment made me go back to the other species to check how we have handled this in the past. Apparently, for the *L. trigynum* case we had included the section despite there only being one reference. Since we have a precedent, let's follow that and let's keep the playwright test like it is. The Reference section is added back in, in bd89d62. 

If you agree with this, I think we can merge and aim for new release next week.

RMCrean:
Sounds good then, lets keep it. 

And yes feel free to merge. I might test the dev cluster today but we can save the release till Monday :)

PR #86, opened 2025-01-23
Update data dockerfile with fix for node.js issue?

brinkdp:
When working on #85 recently, I ran in to an issue that the assembly and track configurations were not passed to `./data/[SPECIES_NAME]/config.json` when running `./scripts/dockermake`. After clearing old `config.json` outfiles, this error was raised: 

```
Using existing local file data/amphiura_filiformis/JAZBNO01.fna.bgz
    throw new ERR_SYSTEM_ERROR(ctx);
          ^

SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
  code: 'ERR_SYSTEM_ERROR',
  info: {
    errno: -2,
    code: 'ENOENT',
    message: 'no such file or directory',
    syscall: 'uv_os_get_passwd'
  },
  errno: [Getter/Setter],
  syscall: [Getter/Setter]
}

Node.js v22.2.0
Using existing local file data/amphiura_filiformis/Afil_fr2py.wn.filt.final.chroms.sorted.gtf
    throw new ERR_SYSTEM_ERROR(ctx);
          ^

SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
  code: 'ERR_SYSTEM_ERROR',
  info: {
    errno: -2,
    code: 'ENOENT',
    message: 'no such file or directory',
    syscall: 'uv_os_get_passwd'
  },
  errno: [Getter/Setter],
  syscall: [Getter/Setter]
}

Node.js v22.2.0
```

Further investigation revealed that this was related to the JBrowse CLI calls in `./scripts/generate_jbrowse_config` (one error for `jbrowse add-assembly` and one for `jbrowse add-track`, to be precise). In fact, all JBrowse CLI commands were failing on my machine when run from the swg-data-builder container. After brainstorming with @RMCrean, we found two GH threads [here](https://github.com/n8n-io/n8n/issues/8664) and [here](https://github.com/GMOD/jbrowse-components/issues/4608) suggesting to add `ENV SHELL=/bin/sh` to the dockerfile. This was successful, and `dockermake` is fully functional on my laptop again with this fix in place.

There are many questions here. Why did this error emerge? Is a MacOS related error? When in time did this happen (I had updated the OS and the *local* install of npm/node.js in December, which might have changed global permissions). Either way, I suspect that it is going to be difficult to reproduce the error on another laptop at will.

More importantly, is this an addition that we want to make to `data.dockerfile`? What does it actually do in the context of node.ja, is it secure, etc? I made the PR to that we can discuss our thoughts on this. I am also happy to consider alternatives. Time is not crucial for this PR, since I have a local fix, and do not rebuild the data-builder very often anyway.


RMCrean:
From memory, I tested it out locally on my laptop and it worked for me, but I can test it now, to be sure. Are you okay to test it out on the dev cluster?

PR #90, opened 2025-02-25
EB-450: Use unique tags for each image build

RMCrean:
This is the first step in the work towards being able to use ArgoCD to update deployments.

- releases are tagged like: v1.2.3
- workflow_dispatch tagged as either (1) custom user input or (2, if no input provided) dev-[commit sha] 
- Push to main tagged as dev-[commit sha]

Next steps:
- Update Kube-swedgene repo to use these image tags (create a new PR which can be merged once this is merged)
- Add genome-portal to ArgoCD instance 

With this setup, order of operations would be 
1.     images built by GH actions
2.    kube-swedgene repo manually updated (update tag in kustomize file, commit and push
3.    sync with ArgoCD.  

Optional addition, could be to automate step 2, either by having the GH actions run make the commit to update the tags for us or trying to use something like https://argocd-image-updater.readthedocs.io/en/stable/ - but that is for after we have the basic ArgoCD setup I think.

### DIFF
--- a/scripts/ci/set_docker_tag
+++ b/scripts/ci/set_docker_tag
@@ -7,11 +7,10 @@ REF_NAME=$2
 SHA=$3
 DOCKER_TAG=$4
 
-if [ -n "$DOCKER_TAG" ]; then
-   tag="$DOCKER_TAG"
-elif [ "$EVENT_NAME" = "release" ]; then
+
+if [ "$EVENT_NAME" = "release" ]; then
     tag="$REF_NAME"
 else
-    tag="dev-$SHA"
+    tag="${DOCKER_TAG:-$SHA}"
 fi
 echo "tag=$tag"
--- a/scripts/ci/set_docker_tag
+++ b/scripts/ci/set_docker_tag
@@ -1,8 +1,17 @@
 #!/bin/sh
-case "$1" in
-    release)
-        tag=prod;;
-    workflow_dispatch)
-        tag="$2";;
-esac
-echo "tag=${tag:-dev}"
+# Note: using getopts here could be a awkward as DOCKER_TAG sometimes exists, sometimes doesn't.
+# DOCKER_TAG occurs if triggered by workflow_dispatch and tag added.
+
+EVENT_NAME=$1
+REF_NAME=$2
+SHA=$3
+DOCKER_TAG=$4
+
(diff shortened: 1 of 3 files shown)


RMCrean:
R.E. the 2nd commit, I added in a "latest" tag to all builds so when building the portal with the` local kustomize overlay `, it can refer to that tag and therefore not worry about being updated. 

Seems to works nicely (tested with workflow_dispatch on this branch), [see e.g. here](https://github.com/ScilifelabDataCentre/genome-portal/pkgs/container/swg-hugo-site)

Now this PR is ready for review I think.

scripts/ci/set_docker_tag
```diff
@@ -1,8 +1,17 @@
...
+SHA=$3
+DOCKER_TAG=$4
+
+if [ -n "$DOCKER_TAG" ]; then
+   tag="$DOCKER_TAG"
+elif [ "$EVENT_NAME" = "release" ]; then
+    tag="$REF_NAME"
+else
+    tag="dev-$SHA"
+fi
```
kwentine, line 16: I think the logic could be simplified a little: if we are doing a release then go for `REF_NAME`; otherwise, take the user-provided tag defaulting to `SHA` if missing.

```suggestion
if [ ${EVENT_NAME} = "release" ]; then
    tag="${REF_NAME}"
else
    tag="${DOCKER_TAG:-dev-$SHA}" # or simply "$SHA" perhaps ?
fi
```

kwentine (approved):
Great! And thank you for taking the trouble to update the bats tests :)
What would you think of having just `<commit sha>` instead of `dev-<commit sha>` ? 
🚀

scripts/ci/set_docker_tag
RMCrean: Good spot, I must make it a part of my process to spot opportunities for using parameter expansion (feels like a recurring them that you spot them in my PRs haha). I've added the change in commit 594de7907ba2d847edc088bb459763ab87e94266 and also swapped to just using $SHA instead of dev-$SHA in that commit.

RMCrean:
> Great! And thank you for taking the trouble to update the bats tests :)
> What would you think of having just <commit sha> instead of dev-<commit sha> ?
> 🚀

Thanks for the review, looking forward to testing this live tomorrow!  
:chipmunk:

PR #89, opened 2025-02-20
EB-427: Remove contact form from the site

RMCrean:
A single commit to make our life easier if we decide to resurrect the contact form. If all looks good I will `rebase and merge` this PR. 

**Commit covers:**
- Remove contact form from the site
- Update contact form text to reflect removal
- Remove playwright tests on contact form
- Remove css for contact form specific styling

**Additional notes**
- Alongside these commits I have removed all records stored by form manager located at https://forms.dc.scilifelab.se/ (I have kept the configuration (as no harm in keeping it) just in case we decide to bring it back later. 
- Once merged I will contact Paul to let him know he doesn't need to notify us of any planned maintenance etc... for form manager.

PR #91, opened 2025-03-05
Remove references to the now deleted contact form

RMCrean:
The contact form was removed in commit c02306023ac98ea8a3f6d9cb3b2b0d562e224266 (see PR #89) but still being refereed to throughout the site and in the `README`, this PR removes those references.

Therefore in the case of reverting the removal of the contact form, one would need to revert the commit(s) in this PR and the commit in PR #89. As in #89, I plan to `rebase and merge` this PR when approved, so there is no merge commit. 

To check the changes more easily I have updated the dev instance of the site to be based on this commit: https://swedgene.scilifelab-2-dev.sys.kth.se/contact/ 

Note: If you do see the contact form when you open the page, hopefully refreshing the page should do the trick - otherwise try to disable your cache and then reload the page. Ideally you wont need to do either of those things though (please let me know if you needed to do that).


brinkdp (approved):
Thanks! I did not find any mentions of the contact form on the dev instance, so that's good. I also checked out the branch and searched the local repo for "contact form", " form ", and "contact". From what I can tell, you seem to have fixed them all. LGTM!

RMCrean:
Awesome, thanks for checking! :)

PR #92, opened 2025-03-25
EB-473: Update glossary for gene annotations

brinkdp:
Based on a suggestion from a user, this PR updates the use of the concept of "annotation" on the Genome Portal to differentiate between _structural_ and _functional_ genome annotation. 

The glossary has been updated with a definition from an external open ontology for "functional annotation", and an in-house definition for "structural annotation". In line with these definitions, the downloads pages for a few species now have been changed to say "Structural annotation of protein-coding genes" instead of "Functional annotation of protein-coding genes".


RMCrean (approved):
Thanks for this, I've built the image locally and checked the changes work as expected. I can confirm that the everything works as expected and that the automatic creation of links for NGBO accession numbers (in `hugo/layouts/glossary/list.html`) works nicely. 

I can't say anything about the definitions themselves of course but LGTM :rocket:

apfuentes:
Thanks for making this clarification and updating the corresponding files and glossary entries. I spotted a minor typo and suggested a change (added word "that"). Other than that, the text and term definition looks good to me :)

PR #95, opened 2025-04-15
Switch to CSI indices for BED and GFF

kwentine:
CSI indices accomodate arbitrarily long chromosomes. Unlike TBI that was inadequate for the spruce, for example. This is an attempt to use CSI indices by default for all BED and GFF files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ define make_index
 	@$(SHELL) scripts/index $<
 endef
 
-$(GFF_INDICES) $(BED_INDICES): %.tbi: %
+$(GFF_INDICES) $(BED_INDICES): %.csi: %
 	$(make_index)
 
 $(FASTA_INDICES): %.fai: %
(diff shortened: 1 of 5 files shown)


brinkdp (approved):
Nice! I found one remaining call for `.tbi` that I changed to `.csi`.

Tested by selectively removing all `.tbi` and config.json from `./data/*/` and `./hugo/static/data/*/` for a clean start. Ran the data build pipeline. Things looked good and checking the About information for each track confirmed that they were calling for `*.csi`. The _Linum_ species GWAS tracks were always manualy crafted, so their defaultSession config.json needed some extra care. 

Looks like everything is working now for our current species :smile:

Curious to test this on dev soon. I wonder if we need to delete the old config.json files on the volume first? I am not sure if `jbrowse add-track` understands that it needs to alter the existing track configs to use TBI instead of CSI?

brinkdp:
Tested this on the dev cluster by merging this branch into `add-green-toad`, which is one of the new assemblies that need CSI indexing. It worked well. I think this can be merged into main.

PR #97, opened 2025-04-30
Extend default session script functionality

brinkdp:
This PR is about adding new functionalities to the support script that configures the JBrowse defaultSession for each species. This PR contains refactoring of the script to adds support for new keys in `config.yml`. These keys do not effect the makefile, and only serve to support this script. The keys are:

- `defaultScaffold: "<string>"`
	- Nested key under `assembly:` . This allows the user to specify the name scaffold that should be displayed by default. For instance: `defaultScaffold: "ENA|CAMGYJ010000002|CAMGYJ010000002.1"`. When `defaultScaffold` is omitted, the script sets the default to the first scaffold of the assembly FASTA (the standard JBrowse behavior).
- `defaultSession: {True, False}`
	- Nested key for use in the list elements in `tracks:`. If set to True, the script will configure the track to be enabled in the defaultSession. Protein-coding genes are an exception, since they will always be configured to be enabled regardless of the inclusion of this key. 
- `GWAS: {True, False}`
	- Nested key for use in the list elements in `tracks:`. Special key intended for tracks that should be displayed with the GWAS plugin (Manhattan plot). When set to True, this key makes the script add the GWAS track to `'.tracks'`; for all other track types at the time of writing, this is done by makefile's call to JBrowse CLI. If the user wants the GWAS track to also be enabled in the defaultSession, the track should also contain `defaultSession: True`; this configures a special case of the `'.defaultSession.views[VIEW_NUMBER].tracks'` that uses the required values from `'.tracks'` to ensure that the internal references are correct (again, this for non-GWAS tracks, this is already harmonised with the output of the JBrowse CLI).
- `scoreColumnGWAS: "<string>"`
	- Finally, for the GWAS track to display properly, this key needs contain the name of the column header in the tabular file that contains the data points that are to be displayed in the Manhattan plot. For instance `scoreColumnGWAS: "TajimaD"`, or `scoreColumnGWAS: "PI"`

 There are also a few test fixtures in `./tests/config/defaultSession_helper/`. I tested them using this pattern:

```
python ./scripts/data_stewardship/configure_default_session.py --yaml ./tests/config/defaultSession_helper/config.yml \
    --out ./tests/config/defaultSession_helper/config.json

SWG_TAG=local ./scripts/dockermake --test SPECIES=defaultSession_helper 
docker stop jb2-defaultSession_helper;SWG_TAG=local ./scripts/browse tests/data/defaultSession_helper
```


RMCrean:
Hey, I think you should feel free to merge this at this time if it will be useful to your other in progress work and we (others) can take a look when you would like to do any refactoring, rather than starting a review process now if you would like to use this code as is.

PR #98, opened 2025-05-05
EB:520 + EB:521: Updates to the species observation map 

RMCrean:
This PR covers 2 related jira tasks:

### EB-520: Investigate potential issues with the Leaflet species map (regarding showing the error message even though it displays correctly - see image from dev cluster on green toad)
Screenshot from 2025-05-05 16-38-11
I've made some changes and refactored the `hugo/assets/js/gbif_map.js` here which _should_ mean this is handled better. For one thing on page load this custom error message is set to display:none so refreshing the page should now clear the error, which is does not seem to do in the current version as seen on the dev cluster which is displaying the green toad branch.    

### EB-521: Make showing the species distribution map optional
I've added some if else logic to the Hugo species intro page template `hugo/layouts/species/species_intro.html ` to be able to handle a case where a species does not have a distribution map to show. In such a case a user would delete the key `.Params.initialZoom` in the species `_index.md` file to activate the alternative path of not showing the observation map. To document this change I have made a commit in the partially-automate-new-species-submissions instead of here because this is where most of that work is going on, see commit fe3e210482a0aad597466f5e13d1d2538b0378a0).

Now the page for a species without an observation map would look something like this: 
Screenshot from 2025-05-05 16-57-50
 

The design/layout for the species pages without a map could probably be improved but that can perhaps be a different Jira task.

### DIFF
--- a/hugo/assets/css/styles.css
+++ b/hugo/assets/css/styles.css
@@ -925,6 +925,12 @@ p + .clipboard {
     background-color: var(--scilife-teal-025);
   }
 
+  .scilife-toast-body {
+    background-color: #FAF9F6;
+  }
+  .scilife-toast-header {
+    background-color: white;
+  }
 }
 
 /* Dark mode specific styles */
@@ -949,6 +955,12 @@ p + .clipboard {
     background-color: var(--scilife-teal);
   }
 
+  .scilife-toast-body {
+    background-color: #212121;
+  }
+  .scilife-toast-header {
+    background-color: #0e1111;
+  }
 
   .scilife-navbar-main li a,
   .scilife-navbar-sub li a,
--- a/hugo/assets/css/styles.css
+++ b/hugo/assets/css/styles.css
@@ -455,6 +455,42 @@ body {
   }
 }
 
+.scilife-intro-img-no-map {
+  width: auto;
+  height: 500px;
+}
+
+@media (max-width: 400px) {
+  .scilife-intro-img-no-map {
+    max-height: 300px;
+  }
+}
+
+@media (min-width: 401px) and (max-width: 576px) {
+  .scilife-intro-img-no-map {
+    max-height: 350px;
+  }
+}
+
+@media (min-width: 577px) and (max-width: 768px) {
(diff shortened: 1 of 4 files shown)


brinkdp (commented):
Thanks! It seem to work as desired when loading up the species pages. But if I play around with the zoom levels on the map, I eventually get this error message that does not go away unless I reload the page:

Screenshot 2025-05-06 at 08 48 21

It happens on all species for me if I click around in the map long enough. Tested on Chrome and Safari, with and without disabling cache.

RMCrean:
@brinkdp, thanks for the report, now fixed!  

This error (`tileerror`) comes from when the browser cancelling an in progress request because the map view changed before the data for all of the previous view was loaded.  See screenshot below for example cancelled requests. 
Screenshot from 2025-05-06 10-12-00

As this is not an error, we now just check on initial load for a `tileerror`, which could come about if the taxon ID used doesn't exist or the GBIF API is down. If during a session the GBIF API went down we wouldn't catch that but that seems very unlikely, would be resolved after refreshing the page.  

I will merge this PR now so we can test it on your new species branches :chipmunk: :rocket:

RMCrean:
See commit 92dcbb3ed4f165a57993e7d9b8e26d5ab42cc3e9 for hopefully the final commit related to this little saga. The comments above the function for `hugo/assets/js/gbif_map.js` have been updated according to this change.

Essentially after this PR was merged, we still weren't handling 204 No Content responses correctly. They cause Leaflet to raise a `tileerror` and you can get these even on initial load of the map (as learned from green toad species). This is now handled by the commit referenced above by checking the header content of the response message and seeing if it is 204. If it is 204 we do not raise an error. 

You would also get a 204 if the taxon id provided does not exist so the checks no longer guard against that (you get a 400 if the request is badly formatted which would raise the error). It was decided that checking the taxon id is beyond the scope of this code. A wrong taxon id is easy to spot by just looking at the map and seeing no data after all...  

I will update here if another commit needs to be added, but hopefully this is my last message on this branch :crossed_fingers:

PR #102, opened 2025-06-05
EB-545: Add a show all species option to home page 

RMCrean:
From internal discussion regarding: 
1. Having a way to click a button to "turn off pagination" and instead show all species, on a single page. 
2. Have a count of the total number of species.

Screenshot from 2025-06-05 14-02-12

In current implementation clicking the button hides the pagination and shows all species. Reloading the page brings back the pagination. Let me know if you don't think this is desired behavior or think we need a way to swap back to having pagination without having to reload the page. (It would be fine to implement.)

### To test current implementation locally you can do:

- Edit the `maxSpeciesPerPage` param in `hugo/config.yaml` to a lower number to trigger pagination, e.g `2`. (Not committed as we of course don't want this). 
- Rebuild and serve image 
     ```
     docker rm -f "genome-portal"; ./scripts/dockerbuild -u -t local -k hugo ; ./scripts/dockerserve -t local
     ```

I also added 2 playwright tests to the home page for it (see 2nd commit)
```
pytest playwright/tests/test_home_page.py --base-url http://localhost:8080 --numprocesses auto
```

### DIFF
--- a/hugo/assets/css/styles.css
+++ b/hugo/assets/css/styles.css
@@ -425,6 +425,19 @@ body {
   border-color: #dee2e6;
 }
 
+.scilife-show-all-btn {
+  color: var(--scilife-teal);
+  background-color: #fff;
+  border-color: var(--scilife-teal);
+}
+
+.scilife-show-all-btn:hover,
+.scilife-show-all-btn:active  {
+  color: #fff;
+  background-color: var(--scilife-teal);
+}
+
+
 
 .scilife-intro-map-img {
   width: auto;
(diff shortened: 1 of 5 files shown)


brinkdp (approved):
Very nice! It looks and handles neat. Today when the number of species are quite few, I think it is fine to use page reload as the way to return to pagination. That said, I think I would personally be in favour of having an option to toggle back to pagination in the future when there are more species. Happy to see this merged in the current state, anyway! 🚀 

What do you think? Would it be easier to implement the toggle now while this is fresh, or better to return to this in the future if/when it becomes needed?

RMCrean:
Awesome, thanks for checking. 
(Based on in person chat we decided to not implement a toggle/swap back to paginated mode on this PR), merging now.

PR #104, opened 2025-06-10
EB-487: Add European Green Toad to the genome portal

brinkdp:
It is time to release the Green Toad to the Genome Portal! I think this branch was already looked at as part of #98, but it would be good if it could through review again to spot things to improve, typos, and similar. Should hopefully go quickly!


RMCrean (approved):
Looks great, nice work :)

PR #105, opened 2025-06-13
EB-562: Fix Genome portal species name alignment problem

RMCrean:
We now have species named that are 2 lines long and the cards no longer align e.g.:
image

As we cannot control the text exactly (e.g. how many lines each section (attribution text, species name, common name) I don't think the problem can be solved perfectly. This PR has 2 potential imperfect "solutions". I vote for the 2nd one.

**Solution 1:** 
Assumes (i.e. text will overflow if not true)
- That the text length of the image attribution will never be more than 3 lines long in the cards.
- That the species title will never be more than 2 lines long. 

**Example output:**
first_commit

With this approach you see the extra space between species scientific and common name even for rows that have no need for it. 

**Solution 2:** 
Assumes (i.e. text will overflow if not true)
- That the text length of the image attribution will never be more than 3 lines long in the cards.
- (No assumptions about species name length).

2nd_commit

Bonus: Example of what would happen if we had an attribution text that was longer than 3 lines.  
2nd_commit_stress_test

So If we take solution 2 as I suggest we should keep in mind things will go wrong if we have an image attribution in the future longer than 3 lines. Would need to modify `scilife-card-image-attrib.height` in `styles.css` accordingly. I think this is unlikely to happen though. 

@apfuentes, @brinkdp do you have any opinions?

### DIFF
--- a/hugo/assets/css/styles.css
+++ b/hugo/assets/css/styles.css
@@ -357,10 +357,11 @@ body {
 }
 .scilife-species-card h2 {
   color: #222;
-  font-size: 1.25rem;
   font-weight: 700;
   letter-spacing: 0.0625rem;
   margin-top: auto;
+  font-size: 1.25rem;
+  height: calc(1.25rem * 2 + 0.5rem); /* this is to get cards of diff size texts to align */
 }
 .scilife-species-card h3 {
   color: var(--scilife-dark-grey);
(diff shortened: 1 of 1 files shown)


RMCrean:
Great @apfuentes, thanks for checking! 

I will merge this PR shortly then, but save it to be released next week (given tomorrow is a day off). 

The max number of characters (including spaces) for 3 lines would be 132.

PR #106, opened 2025-07-04
EB-615: Add pa11y CI GitHub Action

RMCrean:
[Pa11y CI is an accessibility test runner ](https://github.com/pa11y/pa11y-ci). This PR is to add a GH action that runs it. Strategy I'm going for is to run pa11y on PRs to main if the PR touches files that are involved in the creation of the Hugo site.

### DIFF
--- a/.github/pa11y.config
+++ b/.github/pa11y.config
@@ -1,5 +1,7 @@
 {
     "defaults": {
+        "standard": "WCAG2AA",
+        "level": "error",
         "chromeLaunchConfig": {
             "ignoreHTTPSErrors": true,
             "args": ["--disable-dev-shm-usage", "--no-sandbox"]
--- a/.github/pa11y.config
+++ b/.github/pa11y.config
@@ -5,6 +5,6 @@
             "args": ["--disable-dev-shm-usage", "--no-sandbox"]
         },
         "hideElements": "header.scilife-banner h1, #matoOpOut, div.navbar-collapse#navbarSupportedContent"
-    }
+    },
+    "concurrency": 2
 }
-
--- a/.github/pa11y.config
+++ b/.github/pa11y.config
@@ -4,7 +4,7 @@
             "ignoreHTTPSErrors": true,
             "args": ["--disable-dev-shm-usage", "--no-sandbox"]
         },
-        "hideElements": "header.scilife-banner h1, #matoOpOut"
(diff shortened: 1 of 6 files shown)


.github/pa11y.config
```diff
@@ -0,0 +1,10 @@
+{
+    "defaults": {
+        "chromeLaunchConfig": {
```
brinkdp, line 3: I was curious about which accessibility rulesets `pa11y` uses by default and learned from the docs that it is WCAG 2.1 level AA and fail tests only on errors. Sounds good. 

I saw from the some of the [tutorials](https://bitsofco.de/pa11y/) linked on the [pa11y site](https://pa11y.org/tutorials/) that it is possible to explicitly state the ruleset in the config, even if we are just using the defaults. I don't know if being explicit will help for maintainability in the future?

```json
  "defaults": {
    "standard": "WCAG2AA"
    "level": "error"
  }
```

Perhaps it is overkill and might even be an issue if pa11y or the rulesets are updated in the future?

brinkdp (approved):
Great work! Impressive how fast you implemented this! It's rewarding to see that the runner actually found things that could be improved in the site.


> For pathogens, they run a weekly cron job to test the production site. I set it to test on PRs to main if the PR has changes to files that affect the Hugo part of the site. Happy to reconsider if strong opinions. I took the approach I did because I like idea of catching/dealing with the issue during the PR.

Good call, I completely agree that is is enough to check on PR merge. The content on the Genome Portal is quite standarized by now and the pages that get the most updates / new entries are the species pages which bascially follow a template. Cronjobs would be overkill for the Genome Portal at the moment. Checking the genome portal hugo image and only for diffs in `./hugo` sounds good.

I have one small comment/question about the config. Other than that, please go ahead an merge! :rocket:

.github/pa11y.config
RMCrean: Thanks for looking into this, I don't think it hurts to include this, it would be a relatively quick fix if they change the config file format anyway. 

I suppose if we pinned the pa11y version then it would not be a problem at all, but I think taking the latest is best in this case to keep up to date with any new requirements/updates so that they will be automatically tested etc... I have added this now in commit ee5312f71acf440a5227d755fb081b69f134152f

RMCrean:
Thanks for going through and checking, just merged now :chipmunk:

PR #107, opened 2025-09-03
EB-490: Add recipe for trix indexing to genome portal makefile

brinkdp:
This PR is for adding trix indexing to the genome portal backend. [Trix indexes](https://genome.ucsc.edu/goldenPath/help/trix.html) are used by JBrowse 2 to make genes searchable: [`jbrowse text-index`](https://jbrowse.org/jb2/docs/cli/#jbrowse-text-index) looks at all the assemblies in a given JBrowse config file and generates an index that can be used for text searching for genes in the "coordinate box" at the top of a JBrowse view. 

## Details
We have tried trix indexing out in the past by manually putting trix files on the cluster, but these commits adds recipes to the makefile for generating and installing the files. Trix indexing is now added the last step in the `build` recipe (the order is important since the indexing requires that config.json has been finalised). It has been tested locally and on the dev cluster.

The standard behaviour of `jbrowse text-index` is read a `config.json`and to create a subfolder call `trix` in which several index files are stored. The CLI tool also adds an entry in `config.json` that points to these files. The implementation in the `trix-index` recipe makes it so that all compatible tracks associated with each assembly in a `config.json` are indexed. The `*.meta.json`files in the `<SPECIES_NAME>/trix` subdir show which fields that have been indexed for which track. It seems that `jbrowse text-index` supports GFF files with `Name` or `ID` fields (but skips CDS and Exon fields). The `install-trix` recipe ensures that the files are copied to a `trix` subdir in the "install" directory of each species.

There is currently one species (_A. filiformis_) where the protein coding genes are stored in GTF and not GFF. This track was not trix-indexable, despite the `Name` field being present. It is unclear from the  [`jbrowse text-index` manual](https://jbrowse.org/jb2/docs/cli/#jbrowse-text-index) which track formats are supported, so at the moment I assume that GTFs are not compatible with `jbrowse text-index`. The make recipe will create the `trix`subdir for this species too, but it will not populate it and no trix pointers are written to `config.json`, so it is not an issue.

The _P. mnemosyne_ case shows that the indexing works for multi-assembly `config.json`: it is possible to search for genes in either the primary assembly view, or in the mitochondrial assembly view.

## A note on the commit history
There was a bit of trial-and-error to get this to work, which is reflected in the commit history for the first four commits. What I first thought was a working solution (locally) did not work on the cluster. Since I pushed images for the different commits and deployed them on the dev cluster, I decided against cleaning up the commit history with squashing. The working solution was introduced in 8dbcea998cfe07b842aef7ae340963cf66941f33. 

### Commands:
```
./scripts/dockerbuild -u -t local -k data

./scripts/dockermake -t local

(OR specifically for a species: 
./scripts/dockermake -t local SPECIES=linum_tenue
)

```


RMCrean (approved):
Nice, this is a really cool feature! I can confirm this also works for me. Do you want to make a release for this once merged?

PR #110, opened 2025-11-27
SQ-718: Update code for submission form parsing

brinkdp:
This PR is about updating the helper script for adding a new species to correctly parse the latest version of the submission forms. It also contains some updates that were made in the `add-stylops-ater` branch: the main thing there is improved trix index handling for the makefile. Since this branch is likely to be merged before the other branch, I cherry-picked over the relevant commits here. 

## 1. Update `add_new_species` to parse submission forms v1.2 

The new version of the .docx has a pretty new layout that ultimately becomes rendered as tables when converting to markdown. To see the result of the conversion, run this from the root of the repo:

```bash
pandoc --from=docx --to=markdown scripts/add_new_species/tests/fixtures/submission_form_example/01-species-submission_v1.2.docx
```

This complicates the parsing somewhat, as markdown table formatting has to be stripped. That is simple enough for the fields that only have a single column (e.g. species description). For fields such as species name and image attribution that have two columns, the parsing is fairly complex since it needs to handle multi-line entries in both the field and the value cells. "English (Common) name:"  is a case where the field name gets split on multiple lines, for instance. 

All this works on my computer. I just really hope that pandoc renders the tables in the exact same way when this is run on other machines too 😅 

The new version of the spreadsheet was simpler to update. We have dropped a column called `accessionOrDOI` from the spreadsheet since the value is implicit in the url in `doi_link_to_repository` already. This should improve the UX, but this means that we need to parse the urls for the accession number. Parsing logic is now in place for ENA accessions and DOIs, which are the two common ones we see. If we get URLs pointing to other repositories, we will need to update the logic.

To try the updated script with v1.2 example forms:

```bash
python scripts/add_new_species --species-submission-form="scripts/add_new_species/tests/fixtures/submission_form_example/01-species-submission_v1.2.docx" --data-tracks-sheet="scripts/add_new_species/tests/fixtures/submission_form_example/02-Data_Tracks_Form_v1.1.0_fix.xlsx" --species-image="scripts/add_new_species/tests/fixtures/example_images/image_4_3.png" --overwrite
```

## 2. Update how the `makefile` handles trix indexing

The full data builder job that is run when deploying to the cluster was taking a lot of time to index the krill (>20 min) which felt wasteful. Turns out that the cause was the trix indexing (= JBrowse CLI command that makes genes searchable). I thought that I had set up the makefile to not run the indexing if the index files were already present, but this was not working correctly. 

To make a long troubleshooting story short: it turns out that `jbrowse text-index` does not only create the files `<assembly_name>.ix`, `<assembly_name>.ixx`, and `<assembly_name>.json` (the existence of which I thought I was keeping track of in the makefile), but it also adds an array to `config.json`. If the array is present, `jbrowse text-index` will not run the indexing regardless of whether the files exist or not. This does not fit that well with how `make` checks if files exist before deciding to create them. To make matters more complicated, the makefile regenerates `config.json` upon every run to take into account updates from the defaultSession `config.json`. This essentially meant that the indexing was triggered upon every build job since `config.json` would be regenerated without the trix array.

I am a makefile novice and am aware that there are probably more idiomatic ways of doing this. I also needed to add `jq` to `data.dockerfile` to handle parsing of the assmbly name and the trix array. 

For what it is worth, the solution here handles these cases:
1. trix files does not exist in `data/<species_name>/trix`, `data/<species_name>/config.json` does not have trix array.
2. trix files exist in `data/<species_name>/trix`, `data/<species_name>/config.json` has trix array.
3. trix files exist in `data/<species_name>/trix`, `data/<species_name>/config.json` does not have trix array.
4. trix files does not exist in `data/<species_name>/trix`, `data/<species_name>/config.json` does have trix array.

The data builder job that is run on deployment now takes only a few minutes instead of >20 min 🏆 

I think that covers most of the cases we are bound to encounter. 1 and 3 are the cases we would see the most due to new data or the regeneration of `config.json`. Something to keep in mind that always seem to work is to delete the `data/<species_name>/trix` folder and `data/<species_name>/config.json` for a species and then rerun the data builder job.


RMCrean (approved):
Nice work, I can confirm the doc parsing works nicely for me locally too 👍 

The 2 comments are just random praise as I was reviewing! 

Great job with the make file changes! Nice to see the changes to the make file also worked nicely on the dev cluster.

PR #113, opened 2026-03-03
SQ-850: Fix GWAS plugin not rendering

brinkdp:
Hi @RMCrean 

This PR is for fixing the recent error in the GWAS plugin tracks. As I suspected, the error was fixed by bumping JBrowse to the latest version (from 3.2.0 to 4.1.13). I put the JBrowse-bump (fdf5603457eada262fdcf53eb9259d65d9f52c14) on the dev cluster, and also on prod as a hotfix while we wait for this PR to be merged.

I also decided to try to pin the GWAS plugin version so that we can avoid runaway dependency bugs like this. Previously, the plugin was loaded from npm on runtime whenever a user accessed a JBrowse track that used the plugin. Now I changed it so that a pinned version is downloaded during build time, and pointed the config.json files to that path in the docker image instead of the external npm URL. Do you agree that this is a good way to handle this?

Now for a head-scratcher.  Remember that I have had problems building the hugo docker image locally on MacOS since I switched to Rancher Desktop? In order to conveniently test the build time strategy for the plugin, I needed to be able to build local images. Here I admit I had to resort to asking Claude AI for help. Apparently, the combination of alpine for the build image and the GO Garbage Collection settings resulted in a GO memory address handling bug. After iterating on the error with AI, I found that using a debian image (glibc-based) instead of alpine (musl libc-based) for the build step AND turning off GO Garbage Collection fixed the bug. The final stage is still using the nginx image. I deployed to dev from ce90f90 and it seems to work fine. Can you check if you can run `./scripts/dockerbuild -u -t local -k hugo` on this branch on your laptop without problem?

When we are happy with this, I can make a new release. I will also then rebuild the `config.json` files on the clusters so that they point to the GWAS plugin files in the image (at the moment the files on the clusters still point to the remote CDN).

### DIFF
--- a/config/anthophora_plagiata/config.json
+++ b/config/anthophora_plagiata/config.json
@@ -117,7 +117,7 @@
   "plugins": [
     {
       "name": "GWAS",
-      "url": "https://unpkg.com/jbrowse-plugin-gwas/dist/jbrowse-plugin-gwas.umd.production.min.js"
+      "url": "/browser/plugins/jbrowse-plugin-gwas.umd.production.min.js"
     }
   ],
   "tracks": [
--- a/config/anthophora_quadrimaculata/config.json
+++ b/config/anthophora_quadrimaculata/config.json
@@ -117,7 +117,7 @@
   "plugins": [
     {
       "name": "GWAS",
-      "url": "https://unpkg.com/jbrowse-plugin-gwas/dist/jbrowse-plugin-gwas.umd.production.min.js"
+      "url": "/browser/plugins/jbrowse-plugin-gwas.umd.production.min.js"
     }
   ],
   "tracks": [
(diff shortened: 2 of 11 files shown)


RMCrean (approved):
Looks good to me and works for me when I tried it out locally. 

I was a little skeptical at first about using 2 different linux distros in one dockerfile but I had no issues and I suppose as the first 2 stages just build some config files and HTML, CSS and JS etc..., it is okay to mix, especially as it fixes the issue for you on your Mac.

Nice to see the fix implemented on dev cluster instance, making a new release sounds good! 🚀

brinkdp:
Good to hear! Yes, thinking more about it, it sounds a little scary to mix distros like this. Especially since the original error was related to dependencies on different C libraries... I think you are right that the things inherited from the build stage are mainly HTML, CSS, JS, JSON, this should be fine. 

Since the commit at the head of the branch works fine on the dev cluster, I will merge now and make a new release.

PR #123, opened 2026-03-31
SQ-881: Setup API endpoint for GP KPIs

brinkdp:
This PR is about adding an JSON endpoint with the number of species and non-genome assembly data tracks.

As brainstormed in person, this strategy uses a shell script that is run upon hugo docker image build. It counts the species subdirs in `hugo/content/species/*/_index.md` and writes a JSON to `hugo/static/api/metrics.json`. This is then processed by hugo and served by nginx. I decided to use jq for parsing and writing of json files. 

The current idea of the script is to assume that we only have published species in `hugo/content/species/` on prod. 
However, there is currently one exception: `hugo/content/species/clupea_harengus` that is ignored with the `draft: true` in the header. We have not really been using that key since the first public release of the Genome Portal, so maybe we should just delete those draft files and simplify the metrics script to not look for `draft: true`?

## Examples:

```bash
```
then visit http://localhost:8080/api/metrics (or http://localhost:8080/api/metrics.json)

The following non-Docker version also works, but does not have the nginx aliasing so the url needs to end with `.json`

```bash
bash scripts/generate_metrics.sh
cd hugo && hugo serve
```
then visit http://localhost:1313/api/metrics.json


RMCrean (approved):
Nice job, glad it worked out nicely and cool to already see it live on the dev cluster! 🚀 

I think it is fine to remove the draft species and special handling of draft species for this script if that makes things easier.

PR #124, opened 2026-04-15
SQ-890: Containerise the add new species workflow for the genome portal

brinkdp:
This PR is about containerising and documenting the full add_species_workflow so that it can be understood and reliably run and reproduced by users on different computers. The target audience is the Genome Portal staff and advanced users.

Three highlights of this PR:
- Ensure that there are containers for all tools needed, and that they are published to GHCR
- Document full species submission workflow with example commands and fixtures
- Proposal of an addition ot the container image tags: `stable` to refer to latest published release

## 1. Docker container

The new `docker/add_species.dockerfile` bundles all the dependencies and bioinformatics tools needed to run the full species ingestion workflow. 

The [agat](https://agat.readthedocs.io/en/latest/index.html) tool is known to be a little tricky to install due to its many perl dependencies, so I decided to use conda channels for this tool. After our recent discussions, I was intrigued to try `pixi` instead of `mamba`; for this case where we are only installing one tool, I don't notice a big difference in performance with `pixi`, but it was interesting to try. I also decided to pin python 3.11 for this image since there is some awkwardness with the `quast` install command calling `distutils` (depreciated in python 3.12). 

## 2.  Document full species submission workflow 

There is new directory `./species_submission/` that collects the submission form templates and the docs for the species submission. I intend this to be the source-of-truth for the form templates. 

The species ingestion workflow is now documented in `species_submission/how_to_submit_through_pull_request.md`. This is for genome portal staff and for advanced users. It describes the required inputs and the containerised commands that need to be run. 

For convenience, there is also an "I'm feeling lucky" wrapper script `scripts/full_species_ingestion_workflow.sh` that runs all the required steps in the workflow. It has good chances to work for species that do not have quantitative tracks (i.e. tracks where a specific score column in the data file needs to called), as long as the submission forms are correctly filled out.

The `species_submission/local_inputs` dir is intended to be the location to save the local copies of the filled out forms for users that submit with PRs. The dir is not under source control and is mounted by the `add_species.dockerfile` for convenience.

I also did quite a bit of bug fixing and improved error and fallback handling for the scripts as I wrote the docs. Biggest thing is probably that I refactored the old  `scripts/data_stewardship/generate_assembly_and_annotation_statistics.sh` into a Python package for maintainability (to align with the other two packages in the workflow). The submission forms were also updated to better accommodate the data we need to fetch. The form versions are now included inside the forms in a way that allows the add_new_species package to check the version. This is useful since we ask users to rename the files when they email them to us.

## 3. `stable` image tag

To make the commands in the docs for pulling images more convenient for the users, I propose to add a new tag `stable` to our GHCR images named to refer to the latest published release. At first I though of using `main`, but that could be misleading since the main and the latest release might not always be in sync. 

The direct benefit of this is that the user does not need to find the version of the latest release and can just do:

```bash
docker pull ghcr.io/scilifelabdatacentre/swg-add-species:stable && \
docker pull ghcr.io/scilifelabdatacentre/swg-data-builder:stable
```

The tags would then be:
- `latest` = tag referring to the latest image build from any branch. Mutable.
- `vX.Y.Z` = tag referring to a specific semantic version of a release. Intended to be immutable / same tag must never be reused.
- `stable` = tag referring to the latest published release and generated by the release workflow. Mutable.

Current suggestion in `.github/workflows/publish_images.yml` will add the `stable` tag to all release build events. I did not add any logic to make `stable` only allowed in release builds, so it  _can_ be overridden by a manual workflow dispatch if the user sets the tag to `stable`. 

I did not tag any images with `stable` yet. Happy to discuss different naming or strategy for this.

## Commands

This is more or less the commands I describe in the docs. Example with the `latest` tag since the add-species image has not yet been part of a release:

```bash
docker pull ghcr.io/scilifelabdatacentre/swg-add-species:latest && \
docker pull ghcr.io/scilifelabdatacentre/swg-data-builder:latest

cp scripts/add_new_species/tests/fixtures/submission_form_example/01-test-species-submission_v1.3.docx species_submission/local_inputs/ && \
cp scripts/add_new_species/tests/fixtures/submission_form_example/02-test-data-tracks_v1.3.xlsx species_submission/local_inputs/

bash scripts/full_species_ingestion_workflow.sh \
-t latest \
-f /local_inputs/01-test-species-submission_v1.3.docx \
-d /local_inputs/02-test-data-tracks_v1.3.xlsx \
-i /scripts/add_new_species/templates/placeholder_image_4-3_ratio.webp
```


brinkdp:
After discussion in person, we decided to merge this to try out a release to ensure that the `stable` tagging works as expected.

